Issue #11, opened 2022-10-04
Document the expected Sigstore CLI interface


woodruffw:
In order to test the entire matrix, we need a uniform interface that all Sigstore implementations can expose.

The highest common denominator is currently Docker, so whatever interface we produce will need to work with the primitives/mechanisms that Docker exposes (namely environments, volumes, and entrypoints).

In terms of a baseline, we'll need:

* A "sign" interface, which takes an input (e.g. file or env var) and produces the signature (base64 encoded) and certificate (X.509 PEM encoded) corresponding to it
* A "verify" interface, which takes the input, signature, and certificate and returns a true/false value for the signature and certificate's validitity

Additionally, we'll want to parameterize those interfaces a bit, to exercise other functionality:

* For implementations that support it, we'll probably want a "staging" flag that uses Sigstore's staging instances for signing/verification rather than the production instances
* We'll probably want a way to override the OIDC token, to test signing/verification with non-GitHub identities

tetsuo-cpp:
This will be partially achieved in #5, but once it's merged we should take some time to think it through and refine it.

tetsuo-cpp:
This interface will not exist concretely in code as each implementation will be responsible for writing a wrapper for it. However, we should spend some documenting what the expectations are for the entrypoint.

PR #24, opened 2022-10-12
Implement initial GitHub Action

### LINKED ISSUE #1: Implement initial GitHub Action
Now that we have some preliminary Dockerfiles for sigstore-python and cosign, we need to write a test runner to execute across the testing matrix. The script should iterate over all combinations of clients and versions (nightly or stable), build the relevant Docker image and run all tests (this can happen later).

tetsuo-cpp:
Closes #1


test/client.py
```diff
@@ -0,0 +1,60 @@
...
+import subprocess
+
+
+class SigstoreClient:
+    """
+    A wrapper around the Sigstore client under test that provides helpers to
+    access client functionality.
+
+    The `sigstore-conformance` test suite expects that clients expose a CLI that
+    adheres to the protocol outlined at `docs/cli_protocol.md`.
```
tetsuo-cpp, line 11: Let's do this in #11.

.github/workflows/selftest.yml
```diff
@@ -0,0 +1,24 @@
...
+  id-token: write
+
+jobs:
+  selftest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: install sigstore-python
+        run: pip install sigstore
+      - name: conformance test sigstore-python
```
tetsuo-cpp, line 20: This is an example of it in action.

tetsuo-cpp:
Friendly ping @woodruffw and @di. #26 and #32 are based off this branch so we'll need to get this in first.

woodruffw:
Thanks for the ping! Reviewing now.

test/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+
+def pytest_generate_tests(metafunc):
+    """
+    Parametrize each test with the client under test.
+    """
+    entrypoint = metafunc.config.option.entrypoint
+    if "client" in metafunc.fixturenames and entrypoint is not None:
+        metafunc.parametrize("client", [SigstoreClient(entrypoint)])
```
woodruffw, line 30: I think this can just be a `@pytext.fixture` now instead of a dynamic fixture, right?

Something like this:

```python
@pytext.fixture
def client(pytestconfig):
    entrypoint = pytestconfig.getoption("--entrypoint")
    return SigstoreClient(entrypoint)
```

(I think `pytestconfig` is the right config fixture, based on [this](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-pytestconfig))

test/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture(autouse=True)
```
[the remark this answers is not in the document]
tetsuo-cpp: That fixture is what creates the temporary directory for the test to be performed in. It makes it easier to clean up signatures, certificates, etc generated by the client between tests.

test/conftest.py
tetsuo-cpp: [4f2f7fc](https://github.com/trailofbits/sigstore-conformance/pull/24/commits/4f2f7fc4e3f02701dcd3f974883531bd1af93d87)

PR #32, opened 2022-10-13
Document the expected Sigstore CLI interface

### LINKED ISSUE #11: Document the expected Sigstore CLI interface

tetsuo-cpp:
Closes #11


tetsuo-cpp:
I'm still thinking this over, but just thought I'd get your thoughts early.

docs/cli_protocol.md
```diff
@@ -0,0 +1,79 @@
...
+| `--signature` | The path to write the signature to |
+| `--certificate` | The path to write the signing certificate to |
+| `FILE` | The artifact to sign |
+
+### Verify
+
+This subcommand is used to verify the signature for an artifact. The client
+should:
+
+1. Verify that the signing certificate was signed by the Fulcio root
```
tetsuo-cpp, line 57: I think there needs to be some explanation of what the command is expected to do. But I'm a bit worried that this is slowly going to morph into a description of everything that we check in the conformance suite.

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+wrapper that converts the options described in this document to those that the
+client's native CLI accepts.
+
+## Subcommands
+
+This is the set of subcommands that the test CLI must support. Each subcommand
+has a provided syntax, high-level description of what the subcommand is expected
+to do and descriptions of each argument.
+
+To simplify argument parsing, all arguments are required and will **always** be
```
tetsuo-cpp, line 27: My reasons for this are:
- Not all languages make argument parsing convenient. If there's a `sigstore-c` that uses something like `getopt` for argument parsing or even manually, they'll probably appreciate not having to write this code again for conformance testing.
- The cumbersomeness of the CLI doesn't really matter since nobody will invoke it manually.
- I think it's better to avoid optional arguments. If we don't, the conformance test will end up testing things like what flags take precedence over each other, what combinations of flags are allowed, which flags imply others, etc. I think those ideas have value but realistically, I don't think we can use this testing suite to test those things as we can't test the native CLI directly and we'd just be testing the wrapper. We'd also need a lot of prose to document how the wrapper should behave in those cases.

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+5. Write the signature to the disk.
+6. Write the signing certificate to the disk.
+
+```
+ENTRYPOINT sign --signature SIGNATURE_FILE --certificate CERTIFICATE_FILE FILE
+```
+
+| Argument | Description |
+| --- | --- |
+| `--signature` | The path to write the signature to |
```
woodruffw, line 48: Nitpicks: specify these with their metavars, so that it's clear they always take an argument:

```suggestion
| `--signature FILE` | The path to write the signature to |
```

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+7. Verify the Signed Entry Timestamp (SET) of the Rekor log entry.
+8. Verify that the signing certificate was valid at the time of signing.
+
+```
+ENTRYPOINT verify --signature SIGNATURE_FILE --certificate CERTIFICATE_FILE --certificate-email CERTIFICATE_EMAIL --certificate-oidc-issuer CERTIFICATE_OIDC_ISSUER FILE
+```
+
+| Argument | Description |
+| --- | --- |
+| `--signature` | The path to the signature to verify |
```
woodruffw, line 76: Same nitpick as above re: metavars 🙂

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+3. Verify that the expected OIDC issuer is found in the signing certificate's
+   OIDC issuer extension.
+4. Verify that the signature was signed by the signing certificate.
+5. Verify that there is a log entry for the signature on the production Rekor
+   instance.
+6. Verify the Inclusion Proof in the Rekor log entry.
+7. Verify the Signed Entry Timestamp (SET) of the Rekor log entry.
+8. Verify that the signing certificate was valid at the time of signing.
+
+```
```
woodruffw, line 70:
```suggestion
```console
```

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+This subcommand is used to sign an artifact. The client should:
+
+1. Request an ephemeral signing certificate from the production Fulcio instance.
+2. Verify the Signed Certificate Timestamp (SCT) of the signing certificate.
+3. Sign the artifact with this signing certificate and produce a signature.
+4. Upload this signature to the production Rekor instance.
+5. Write the signature to the disk.
+6. Write the signing certificate to the disk.
+
+```
```
woodruffw, line 42:
```suggestion
```console
```

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
...
+ENTRYPOINT sign --signature SIGNATURE_FILE --certificate CERTIFICATE_FILE FILE
```
woodruffw, line 43: It might be worth denoting `ENTRYPOINT` with braces or a dollar sign here, to emphasize that it's an intended substitution. That's obvious to you and me, but a visual hint might help other readers 🙂 

e.g.:

```suggestion
${ENTRYPOINT} sign --signature SIGNATURE_FILE --certificate CERTIFICATE_FILE FILE
```

docs/cli_protocol.md
```diff
@@ -0,0 +1,80 @@
+Conformance CLI Protocol
+========================
+
+While all Sigstore clients share some core functionality, each of them have
+their own unique features and idiosyncracies that make them incompatible with
```
woodruffw, line 5: Tiny nit, just to emphasize that the CLIs are incompatible, not the underlying functionality (which is why we're testing it):

```suggestion
their own unique features and idiosyncracies that make their command lines incompatible with
```

woodruffw:
This looks good to me overall! I left a handful of small nits, but the expected interface looks reasonable (and not burdensome for other implementations to comply with).

docs/cli_protocol.md
tetsuo-cpp: True! Good catch.

docs/cli_protocol.md
tetsuo-cpp: [3e5667e](https://github.com/trailofbits/sigstore-conformance/pull/32/commits/3e5667ed037f70d8208eb952c477d347acf40cdb)

docs/cli_protocol.md
tetsuo-cpp: [3e5667e](https://github.com/trailofbits/sigstore-conformance/pull/32/commits/3e5667ed037f70d8208eb952c477d347acf40cdb)

sigstore-python-conformance
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
+"""
+A wrapper to convert `sigstore-conformance` CLI protocol invocations to match `sigstore-python`.
```
tetsuo-cpp, line 4: This can eventually go in the `sigstore-python` repository, but for now it's helpful to keep the self test working.

docs/cli_protocol.md
```diff
@@ -0,0 +1,78 @@
...
+   OIDC issuer extension.
+3. Verify that the signature was signed by the signing certificate.
+4. Verify that there is a log entry for the signature on the production Rekor
+   instance.
+5. Verify the Inclusion Proof in the Rekor log entry.
+6. Verify the Signed Entry Timestamp (SET) of the Rekor log entry.
+7. Verify that the signing certificate was valid at the time of signing.
+
+```console
+${ENTRYPOINT} verify --signature FILE --certificate FILE --certificate-oidc-issuer URL FILE
```
tetsuo-cpp, line 70: I unfortunately had to remove the certificate email verification. From what I can tell, there doesn't seem to be any programmatic way to figure out what the SAN will be.

It seems to be roughly:
```
${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/${GITHUB_WORKFLOW}@${GITHUB_REF}
```
But unfortunately, `GITHUB_WORKFLOW` only evaluates to the full path (which we need) if the workflow doesn't have a `name` attribute. Otherwise, it evaluates to the name.

docs/cli_protocol.md
```diff
@@ -0,0 +1,78 @@
...
+${ENTRYPOINT} sign --signature FILE --certificate FILE FILE
+```
+
+| Option | Description |
+| --- | --- |
+| `--signature FILE` | The path to write the signature to |
+| `--certificate FILE` | The path to write the signing certificate to |
+| `FILE` | The artifact to sign |
+
+### Verify
```
tetsuo-cpp, line 53: I think we'll eventually need a `Verify Bundle` subcommand, but for now I think this covers what we care about.

docs/cli_protocol.md
woodruffw: Hmm -- we can verify this when we begin to use it on `sigstore-python`, but I _think_ we can get around this by mandating that every conformance testsuite user uses the exact same workflow filename/`name` (and providing an org-level template that assists with that).

woodruffw: In other words, if we need `GITHUB_WORKFLOW` to always be `sigstore-confirmance`, we should document that and fail the action here if it's anything else 🙂

docs/cli_protocol.md
woodruffw: I agree with this rationale.

docs/cli_protocol.md
woodruffw: Rather than a separate `verify bundle` subcommand, I think we'll probably end up with CLIs that look something like this:

```
sigstore sign foo # produces foo.sigstore
sigstore verify foo # loads foo.sigstore
```

In other words, the current expected interfaces will be deprecated and replaced with bundle-only ones.

docs/cli_protocol.md
woodruffw: Yeah, I think we should remove this description. Instead, we should enforce all of these properties within the conformance suite itself and say that the subcommand must exit with just a success or failure status corresponding to the expected behavior.

In other words: we pass the verification materials in, and confirm that the status code matches our expectations of how the command behaves.

sigstore-python-conformance
```diff
@@ -0,0 +1,19 @@
...
+"""
+
+import subprocess
+import sys
+
+# Trim the script name.
+fixed_args = sys.argv[1:]
+
+# Replace incompatible flags.
+fixed_args = [arg if arg != "--certificate-oidc-issuer" else "--cert-oidc-issuer" for arg in fixed_args]
```
woodruffw, line 14: Is `--certificate-oidc-issuer` the option that `cosign` uses? IMO we should use `--cert-oidc-issuer` instead and avoid the wrapper if we can, assuming that it's compatible with `cosign`.

tetsuo-cpp: Ok, maybe I'll make the flags as `cosign` compatible as possible. So use `--cert` instead of `--certificate` too.

docs/cli_protocol.md
tetsuo-cpp: The problem is that the SAN doesn't contain the workflow name, it contains the path to the workflow file (regardless of whether it has a `name` or not).

So in order for this to work, we'd not only have to mandate the same workflow filename, but also the same path in the repo (`.github/workflows/conformance.yml` maybe?).

woodruffw: I _think_ that'll be okay -- workflows can only be under `.github/workflows` anyways, so mandating the workflow filename is not too onerous.

sigstore-python-conformance
tetsuo-cpp: My reasoning for doing it like this is that pretty much every client is going to need to write their own wrapper (including `cosign` since they use `sign-blob` and `verify-blob`). So the only client that can avoid writing a wrapper if I do it as you describe would be `sigstore-python`.

I figured that if most clients need to write their own wrapper, I should stick to conventions which I think make the most sense for this purpose which I believe are:
- Flags in the same order every time.
- Flag names are verbose and descriptive as possible since they don't need to be manually invoked.

docs/cli_protocol.md
tetsuo-cpp: Ok, cool. I think that's reasonable.

docs/cli_protocol.md
tetsuo-cpp: [ac64a54](https://github.com/trailofbits/sigstore-conformance/pull/32/commits/ac64a54bf402bc026449d586fe310aa8603d733e)

tetsuo-cpp:
The self-test is currently failing due to https://github.com/sigstore/sigstore-python/pull/288 so we'll need to get that in first.

sigstore-python-conformance
woodruffw: That's reasonable! I'm okay with leaving it as the longer option.

docs/cli_protocol.md
woodruffw: I'll make a follow-up issue to track this.

woodruffw: https://github.com/trailofbits/sigstore-conformance/issues/51

Issue #30, opened 2022-10-12
Verify: Validation test case enumeration


asraa:
Hi all!

I was curious about starting a list of test-cases for validation failure modes during verification. Some of these are difficult because it requires tampering with the Rekor responses, which won't be solved until https://github.com/trailofbits/sigstore-conformance/issues/25 is done.

I wanted to take a stab at creating a list of the failure modes... I'm sure I am missing some.

<table>
  <tr>
   <td><strong>Failure Mode</strong>
   </td>
   <td><strong>Testcase(s)</strong>
   </td>
  </tr>
  <tr>
   <td>Signature Validation Failure
   </td>
   <td>
<ol>

<li>Mismatched artifact against a given signing certificate and signature.

<li>Mismatched signature against a given artifact and signing certificate.

<li>Mismatch signing certificate against a given artifact and certificate.

<li>Empty signature.

<li>Empty signing certificate.

<li>Empty artifact.
</li>
</ol>
   </td>
  </tr>
  <tr>
   <td>Certificate Validation Failure
   </td>
   <td>
<ol>

<li>Invalid Certificate chain: not issued by configured root CA

<li>Certificate issued after Fulcio Root CA expiration  [MOCK]
</li>
</ol>
   </td>
  </tr>
  <tr>
   <td>Short-lived certificate signature validation failure
   </td>
   <td>
<ol>

<li>Bypass transparency log upload: no transparency log entry returned.

<li>Invalid transparency log entry returned: mismatch with entry. [MOCK]

<li>Invalid transparency log entry returned: incorrectly signed entry timestamp [MOCK]

<li>Invalid transparency log entry returning: invalid inclusion proof  [MOCK]

<li>Invalid transparency log entry returning: invalid signed root hash  [MOCK]

<li>Integrated time of log entry was past certificate validity period.
</li>
</ol>
   </td>
  </tr>
  <tr>
   <td>Offline Rekor proofs or Bundles
   </td>
   <td>
<ol>

<li>Invalid bundle: Invalid signed entry timestamp 

<li>Invalid bundle: does not match provided signature, artifact, and certificate.

<li>Integrated time of offline log entry was past certificate validity
</li>
</ol>
   </td>
  </tr>
</table>

woodruffw:
That's a great starting list, thanks for enumerating and collecting it!

I think some of these will be captured in the conformance suite, but it'll be great to have them as standalone tests within sigstore-python as well.

[...]

PR #50, opened 2022-11-08
test: Add more verification tests

tetsuo-cpp:
Partially addresses #30


test/test_signature_verify.py
```diff
@@ -0,0 +1,107 @@
...
+    Tests that verification fails with mismatching artifacts, certificates and
+    signatures.
+    """
+    a_artifact_path = Path("a.txt")
+    a_certificate_path = Path("a.txt.crt")
+    a_signature_path = Path("a.txt.sig")
+
+    assert a_artifact_path.exists()
+    assert not a_certificate_path.exists()
+    assert not a_signature_path.exists()
```
woodruffw, line 76: I might be misreading this, but is this test implicitly dependent on `test_verify_empty` running first?

If so, I don't think we want that -- `pytest` doesn't actually guarantee the order in which tests are run, and plugins like `xdist` could conceivably change that order if we decide to use them to speed the CI up.

Instead, I think we want every test to be self-contained, with pre-baked assets wherever possible. For tests that do create assets, they should put them under a temporary directory that `pytest` itself manages, so that we get auto-cleanup after each run.

woodruffw: Specifically, we probably want the `tmp_path` fixture: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

That'll give a temporary directory for each test, which we can then place individual files within.

test/test_signature_verify.py
```diff
@@ -0,0 +1,107 @@
...
+    # Sign the artifact.
+    client.sign(artifact_path, certificate_path, signature_path)
+
+    assert certificate_path.exists()
+    assert signature_path.exists()
+
+    # Write a blank temporary file.
+    empty_path = Path("empty.txt")
+    empty_path.open("w").close()
+    _swap_files(artifact_path, empty_path)
```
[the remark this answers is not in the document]
woodruffw: Actually, taking a step back a bit: do we need this helper? IMO it makes the assertions a little more confusing below, since it obscures the fact that we're replacing the underlying contents by keeping the variable names the same.

test/test_signature_verify.py
tetsuo-cpp: This doesn't rely on the empty test.

I already have a fixture for copying the artifacts into a workspace directory that gets blown away on teardown. That `empty_path` code in the other test is a way of clearing out the artifact, signature and certificate. So I swap the blank file with the signature, try to verify, swap the signature back in, swap the blank file for the certificate, etc.

test/test_signature_verify.py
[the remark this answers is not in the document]
tetsuo-cpp: Oops, no idea what I was thinking. I don't need to swap anything because the path of the materials, doesn't really matter.

Sorry, I'll fix that up now.

test/test_signature_verify.py
woodruffw: Whoops, my bad! I forgot that `workspace` is marked as `autouse`. Maybe we should make that fixture explicit, and then omit it in the other test so that we don't have to do the manual cleanup?

test/test_signature_verify.py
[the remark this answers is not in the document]
tetsuo-cpp: [f85c42d](https://github.com/trailofbits/sigstore-conformance/pull/50/commits/f85c42d8bf50832050a9fce8cd47e44c47e52b26)

test/test_signature_verify.py
tetsuo-cpp: Ok, how's this? I haven't made the fixture explicit but I've fixed up the other test so it shouldn't be as confusing.

There was no cleanup going on in that test. I just had an extended brainfart (sorry!) that made me think that the verification materials needed to be in the exact same spot every time (so I was replacing each of the materials with a blank file).

woodruffw: This is much easier to follow, thanks!

woodruffw (approved):

PR #56, opened 2022-12-09
[DO NOT MERGE]: Move conformance workflow to `pull_request_target` trigger

tetsuo-cpp:
I'm doing some testing to figure out what identity gets used when we run on the `pull_request_target` trigger.


tetsuo-cpp:
Ah, I think in order to really test this, I need to make a second fork and then point a PR to my initial fork. That's because the `pull_request_target` needs to be already be in the target branch to run in Actions.

PR #57, opened 2022-12-09
README: Elaborate on what we aim to achieve with CLI testing

tetsuo-cpp:
I've added a bit more detail to the README to capture what we discussed in https://github.com/sigstore/cosign/issues/2434 in regards to what we aim to achieve with CLI testing as opposed a vector based approach.

CC: @znewman01

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,29 @@ sigstore-conformance
 
 `sigstore-conformance` is a conformance testing suite for Sigstore clients.
 
+This suite provides a high-level view of client behaviour as a whole and sets
+out to answer questions such as:
+- What files are written to the disk by the client in what situations?
+- Does the client fail when given a signing certificate that isn't signed by
+  the Fulcio root CA during the signing workflow?
+- Does the client fail when given an invalid inclusion proof from Rekor during
+  the verification workflow?
+- Does the client fail when given an invalid signed certificate timestamp as
+  part of the Fulcio response in the signing workflow?
+- etc
+
+Over time, `sigstore-conformance` aims to become an informal specification that
+describes how Sigstore clients are expected to behave.
+
+Some general testing principles for this suite are:
+- *Tests should be "workflow" focused.* This testing suite is not about fuzzing
+  every possible input to the client CLI or achieving code coverage.
+- *Tests should exercise the entire client end-to-end rather than individual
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,29 @@ sigstore-conformance
...
+- What files are written to the disk by the client in what situations?
+- Does the client fail when given a signing certificate that isn't signed by
+  the Fulcio root CA during the signing workflow?
+- Does the client fail when given an invalid inclusion proof from Rekor during
+  the verification workflow?
+- Does the client fail when given an invalid signed certificate timestamp as
+  part of the Fulcio response in the signing workflow?
+- etc
+
+Over time, `sigstore-conformance` aims to become an informal specification that
```
znewman01, line 22: How does this related to the "formal" specification for Sigstore clients?

https://github.com/sigstore/architecture-docs

README.md
```diff
@@ -8,6 +8,29 @@ sigstore-conformance
 
 `sigstore-conformance` is a conformance testing suite for Sigstore clients.
 
+This suite provides a high-level view of client behaviour as a whole and sets
+out to answer questions such as:
+- What files are written to the disk by the client in what situations?
```
znewman01, line 13: Hm....I don't actually view "writing files to disk" as part of the remit of most Sigstore clients. This seems to conflate "Sigstore client" with "Sigstore CLI app," and a quite specific one at that.

I actually *don't* expect that we would have a proliferation of Sigstore CLI apps. I'd expect there to be really only one (basically, an improved version of Cosign without the OCI baggage). Then, most clients of Sigstore would be application-specific (in the Python case, built into Pip, eventually).

(The remaining points all seem reasonable.)

znewman01 (approved):
Hey, thanks so much for writing this up! Very helpful.

Couple clarifying questions but LGTM overall

README.md
tetsuo-cpp: I wasn't aware of that.

Maybe I can try something like:

> An official Sigstore client specification is being worked on at the moment as part of the [Sigstore Architecture Documentation](https://github.com/sigstore/architecture-docs). Once it's complete, `sigstore-conformance` aims to be able to test a client's adherence to the specification.

znewman01: SGTM

Issue #65, opened 2023-04-03
Add tests for bundle support


tetsuo-cpp:
Once #51 is done, we should look at exercising bundle support. We should write some basic unit tests exercising that bundles are created correctly and can be verified with the same client.

loosebazooka:
How are we specifying bundles? Currently the cli interface (interface interface?) in sigstore-java allows specifying either sig/cert output files or a bundle output file. But I think from our last slack discussion, it wasn't clear if the python cli could enforce those constraints?

tetsuo-cpp:
@tnytown will figure that out as part of #51 first. But I think we can/should make the spec match `sigstore-java`'s CLI behaviour.

In the current spec, there's an emphasis on guaranteeing that flags are in order and are non-optional, etc for ease of argument parsing. I think we should mostly get rid of that emphasis and allow:
```
<client> verify --signature foo.sig --certificate foo.sig foo.txt
```
or
```
<client> verify --bundle foo.sigstore too.txt
```

That way, the CLI isn't **only** useful for conformance testing but is more ergonomic for general use like what you're using it for with `sigstore-java`. Let me know whether that answers your question.

woodruffw:
#66 contains some concrete examples of test cases we'll want to exercise with this functionality.

PR #67, opened 2023-04-18
Bundle creation and consumption specs

### LINKED ISSUE #51: CLI interface: specify bundle creation and consumption
This is a follow-on to #32 and #11: once Sigstore bundles are stabilized and supported in a few different clients, we'll need to update the CLI specification here to include them.

tnytown:
Specifies bundle creation and consumption standards by:

- adding `sign-bundle` and `verify-bundle` subcommands;
- adding the `--bundle` flag that specifies the bundle to output or verify from.

I'm unsure of if we want to have separate `sign-bundle` and `verify-bundle` subcommands, as I mentioned in https://github.com/sigstore/sigstore-conformance/issues/65#issuecomment-1513638198. I can remove these subcommands and make some flags optional if we feel that approach is better.

Resolves #51.

### DIFF
--- a/docs/cli_protocol.md
+++ b/docs/cli_protocol.md
@@ -58,8 +58,7 @@ ${ENTRYPOINT} sign-bundle --bundle FILE FILE
 #### Signature and certificate flow
 
 ```console
-${ENTRYPOINT} verify --signature FILE --certificate FILE
---certificate-identity IDENTITY --certificate-oidc-issuer URL FILE
+${ENTRYPOINT} verify --signature FILE --certificate FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL FILE
 ```
 
 | Option | Description |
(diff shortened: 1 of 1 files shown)


tetsuo-cpp:
I understand where you're coming from regarding the flags. I guess my only concern is that `sigstore-java` are using their CLI not **just** for conformance testing but for some internal CI stuff to sign/verify things (from my memory). So this approach (which I'd say is a bit less ergonomic) will affect them too.

@loosebazooka Would it be ok if we switch the `sigstore-java` CLI to use `sign-bundle` and `verify-bundle` as separate subcommands rather than a `--bundle` flag? I'm happy to put a PR together, it's more about whether you intend to use the CLI more extensively.

tetsuo-cpp (approved):
LGTM! Provided that the `sigstore-java` folks are happy with this approach, let's go with this.

tetsuo-cpp:
@loosebazooka Awesome, thanks for explaining.

PR #68, opened 2023-04-19
workflows: pin python version to 3.11

tnytown:
The changeset for #65 depends on Python 3.8 features, which causes issues with the CI workflow pinned on 3.7: https://github.com/trail-of-forks/sigstore-conformance/actions/runs/4747768731/jobs/8433107457

To fix that issue, this changeset pins both the conformance testing workflow and the CI workflow to a consistent version of Python.


tetsuo-cpp (approved):
LGTM!

PR #69, opened 2023-04-19
Basic bundle tests

### LINKED ISSUE #65: Add tests for bundle support

tnytown:
Adapts existing tests to cover the bundle flow (empty materials, mismatched materials, ...) To this effect, extends `SigstoreClient` to support the `*-bundle` subcommands.

Depends on #51.
Resolves #65.

### DIFF
--- a/test/client.py
+++ b/test/client.py
@@ -15,7 +15,7 @@ CERTIFICATE_OIDC_ISSUER = "https://token.actions.githubusercontent.com"
 class VerificationMaterials:
     """
     A wrapper around verification materials. Materials can be either bundles
-    or signatures and certificates.
+    or detached pairs of signatures and certificates.
     """
 
     @classmethod
--- a/test/client.py
+++ b/test/client.py
@@ -142,7 +142,9 @@ class SigstoreClient:
         )
 
     @sign.register
-    def _sign_for_bundle(self, materials: BundleMaterials, artifact: os.PathLike) -> None:
+    def _sign_for_bundle(
+        self, materials: BundleMaterials, artifact: os.PathLike
+    ) -> None:
         """
         Sign an artifact with the Sigstore client, producing a bundle.
 
@@ -204,7 +206,9 @@ class SigstoreClient:
         )
 
     @verify.register
-    def _verify_for_bundle(self, materials: BundleMaterials, artifact: os.PathLike) -> None:
+    def _verify_for_bundle(
+        self, materials: BundleMaterials, artifact: os.PathLike
+    ) -> None:
         """
         Verify an artifact given a bundle with the Sigstore client.
 
--- a/test/client.py
+++ b/test/client.py
@@ -79,6 +79,10 @@ class SigstoreClient:
 
     The `sigstore-conformance` test suite expects that clients expose a CLI that
     adheres to the protocol outlined at `docs/cli_protocol.md`.
+
+    The `sign` and `verify` methods are dispatched over the two flows that clients
+    should support: signature/certificate and bundle. The overloads of those
+    methods should not be called directly.
     """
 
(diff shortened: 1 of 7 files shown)


test/client.py
```diff
@@ -38,45 +102,86 @@ def run(self, *args) -> None:
...
+
+    @verify.register
+    def _(self, materials: BundleMaterials, artifact: os.PathLike) -> None:
+        self.run(
+            "verify-bundle",
+            "--bundle",
+            materials.bundle,
             "--certificate-identity",
             CERTIFICATE_IDENTITY,
             "--certificate-oidc-issuer",
```
woodruffw, line 187: Clever use of `singledispatch` 🙂 

I'm not opposed to this use of `functools`, but we should document here with the expectation that other maintainers of other (non-Python) clients will want to make changes in the future. As such, let's add some docstrings to each of these methods (plus top-level module and class docstrings) that explain what we're doing and why.

test/conftest.py
```diff
@@ -38,6 +33,27 @@ def client(pytestconfig):
...
+
+    return _make_materials_by_type
+
+
+@pytest.fixture(params=[BundleMaterials, SignatureCertificateMaterials])
+def make_materials(request, make_materials_by_type):
+    def _curry(input_name: str):
+        return make_materials_by_type(input_name, request.param)
+
+    return _curry
```
woodruffw, line 54:
```suggestion
    def _make_materials(input_name: str):
        return make_materials_by_type(input_name, request.param)

    return _make_materials
```

Let's avoid CS theory names when possible; a reader doesn't need to know what `curry` means to see that we're doing function specialization/partial evaluation 🙂

test/test_signature_verify.py
```diff
@@ -3,88 +3,114 @@
 
 import pytest  # type: ignore
 
-from .client import SigstoreClient
+from .client import SignatureCertificateMaterials, SigstoreClient
 
 
-def test_verify_empty(client: SigstoreClient) -> None:
+def test_verify_empty(client: SigstoreClient, make_materials) -> None:
```
woodruffw, line 9: If I'm reading correctly, this now means that all of our tests with `make_materials` are parametric over both bundles and sig/cert pairs, correct?

If so, very nice!

woodruffw (commented):
LGTM overall, a couple of notes about naming and documentation.

test/client.py
```diff
@@ -8,13 +12,77 @@
 CERTIFICATE_OIDC_ISSUER = "https://token.actions.githubusercontent.com"
 
 
+class VerificationMaterials:
+    """
+    A wrapper around verification materials. Materials can be either bundles
+    or signatures and certificates.
```
woodruffw, line 18: Nit: clarity:

```suggestion
    or detached pairs of signatures and certificates.
```

test/conftest.py
```diff
@@ -38,6 +33,27 @@ def client(pytestconfig):
     return SigstoreClient(entrypoint, identity_token)
 
 
+@pytest.fixture
+def make_materials_by_type():
```
woodruffw, line 37: Sorry, missed this in the last round: let's add a docstring on this fixture and the one below explaining the fixture/parametric behavior.

test/test_signature_verify.py
woodruffw: I'm good with it! IMO magic is justified on the pytest side so long as it (1) makes parametric testing easier, and (2) doesn't surface to users in confusing ways. This is _potentially_ confusing, but IMO with documentation (see the other comment) it's acceptable.

woodruffw (approved):
LGTM; great work @tnytown!

PR #72, opened 2023-04-24
Give the OIDC broadcaster a scary name

woodruffw:
To emphasize that it must not be trusted.

CC @tetsuo-cpp @di


haydentherapper (commented):
Could we disable this workflow for now until token encryption is added?

woodruffw:
> Could we disable this workflow for now until token encryption is added?

Yep. I need to merge it before disabling, though, since the option to disable it won't appear until it's on the "main" branch 🙂

haydentherapper (approved):

woodruffw:
Disabled!

PR #74, opened 2023-04-26
treewide: Use OIDC beacon instead of ambient credentials

tetsuo-cpp:


tetsuo-cpp:
I still had to add the "safe to test" label since these changes aren't in `main` yet, but we shouldn't have to after this is merged.

action.py
```diff
@@ -19,6 +26,92 @@
...
+        raise OidcTokenError(
+            f"Found unexpected number of artifacts on OIDC beacon run: {artifacts}"
+        )
+
+    oidc_artifact = artifacts[0]
+    if oidc_artifact["name"] != "oidc-token":
+        raise OidcTokenError(
+            f"Found unexpected artifact on OIDC beacon run: {oidc_artifact['name']}"
+        )
+    artifact_id = oidc_artifact["id"]
```
tetsuo-cpp, line 101: I was thinking a bit about what parts of the JSON shape I should check for explicitly and which ones I should just let fail with index errors, etc.

I decided not add checks for anything that is guaranteed by the endpoint (for example, if we get a reply here, the API guarantees we have `id` and `name` fields). But anything that we believe to be guaranteed by our OIDC beacon (that we have at least one workflow, the correct artifact, etc) should be checked for.

test/client.py
```diff
@@ -51,7 +49,14 @@ def sign(
         `certificate` is the path to write the signing certificate to.
         """
         self.run(
-            "sign", "--signature", signature, "--certificate", certificate, artifact
+            "sign",
+            "--identity-token",
```
woodruffw, line 53: This needs to be added to the `cli-protocol.md`, right? 🙂

test/client.py
```diff
@@ -2,11 +2,8 @@
 import subprocess
 
 CERTIFICATE_IDENTITY = (
-    os.environ["GITHUB_SERVER_URL"]
-    + "/"
-    + os.environ["GITHUB_REPOSITORY"]
-    + "/.github/workflows/conformance.yml@"
-    + os.environ["GITHUB_REF"]
+    "https://github.com/sigstore-conformance/extremely-dangerous-oidc-broadcaster/.github/"
+    "workflows/oidc.yml@refs/heads/main"
```
woodruffw, line 6: Nice!

action.py
```diff
@@ -58,6 +151,14 @@ def _fatal_help(msg):
 if entrypoint:
     sigstore_conformance_args.extend(["--entrypoint", entrypoint])
 
+gh_token = os.getenv("GHA_SIGSTORE_CONFORMANCE_GITHUB_TOKEN")
+if gh_token:
+    try:
+        oidc_token = _get_oidc_token(gh_token)
+    except OidcTokenError as e:
+        _fatal_help(f"Could not retrieve OIDC token: {str(e)}")
+    sigstore_conformance_args.extend(["--identity-token", oidc_token])
```
woodruffw, line 160: I might be missing it, but what's the fallthrough case here? Given that the GitHub token should always be present, maybe we could drop the `if` block entirely?

action.yml
```diff
@@ -34,5 +27,6 @@ runs:
         ${{ github.action_path }}/action.py
       env:
         GHA_SIGSTORE_CONFORMANCE_ENTRYPOINT: "${{ inputs.entrypoint }}"
+        GHA_SIGSTORE_GITHUB_TOKEN: "${{ github.token }}"
```
woodruffw, line 30: Rather than plumbing through our own environment, could we use `GITHUB_TOKEN` directly? That should be exactly the same as the `${{ github.token }}` expansion 🙂

action.py
```diff
@@ -19,6 +26,92 @@
...
+        url=_OIDC_BEACON_API_URL + f"/artifacts/{artifact_id}/zip",
+        headers=headers,
+    )
+    resp.raise_for_status()
+
+    artifact_zip = ZipFile(BytesIO(resp.content))
+    artifact_file = artifact_zip.open("oidc-token.txt")
+
+    # NOTE(alex): Strip newline.
+    return artifact_file.read().decode()[:-1]
```
woodruffw, line 114:
```suggestion
    return artifact_file.read().decode().rstrip()
```

...perhaps?

woodruffw, line 114: I believe `ZipFile` can be a context manager, so let's use it like that to avoid a resource leak:

```python
with ZipFile(...) as artifact_zip:
    ...
```

action.py
```diff
@@ -19,6 +26,92 @@
...
+    "actions"
+)
+_OIDC_BEACON_WORKFLOW_ID = 55399612
+
+
+class OidcTokenError(Exception):
+    pass
+
+
+def _get_oidc_token(gh_token: str) -> str:
```
woodruffw, line 40: Rather than passing in the token as a parameter here, what do you think about making this parameterless and having it do the `getenv("GITHUB_TOKEN")` directly? IMO that would more directly convey the operation here, and wouldn't imply that we're more flexible about tokens than we actually are 🙂

action.py
```diff
@@ -58,6 +151,14 @@ def _fatal_help(msg):
 if entrypoint:
     sigstore_conformance_args.extend(["--entrypoint", entrypoint])
 
+gh_token = os.getenv("GHA_SIGSTORE_CONFORMANCE_GITHUB_TOKEN")
```
woodruffw, line 154: Same comment as https://github.com/sigstore/sigstore-conformance/pull/74/files#r1177391839 -- maybe this?

```suggestion
gh_token = os.getenv("GITHUB_TOKEN")
```

action.py
woodruffw: That sounds good to me for now! IMO we can defer some of the more precise error handling here until we roll this out to conformance suite users, and see whether they experience or are confused by any errors that happen.

woodruffw (commented):
A couple of nits, but otherwise LGTM! I'm excited to get this merged 🙂

action.py
tetsuo-cpp: [5e2a0ef](https://github.com/sigstore/sigstore-conformance/pull/74/commits/5e2a0ef2ef2e27dfcfcc13b9a018988267bc92ab)

tetsuo-cpp: [c337c79](https://github.com/sigstore/sigstore-conformance/pull/74/commits/c337c795d6e64f0cf8ecc4c55cd7469c27c7755c)

action.py
tetsuo-cpp: [79b2b1e](https://github.com/sigstore/sigstore-conformance/pull/74/commits/79b2b1e721b310dfe2bc4f0dc1b5e4001cf2efc3)

action.py
tetsuo-cpp: [79b2b1e](https://github.com/sigstore/sigstore-conformance/pull/74/commits/79b2b1e721b310dfe2bc4f0dc1b5e4001cf2efc3)

action.yml
tetsuo-cpp: [79b2b1e](https://github.com/sigstore/sigstore-conformance/pull/74/commits/79b2b1e721b310dfe2bc4f0dc1b5e4001cf2efc3)

action.py
tetsuo-cpp: [79b2b1e](https://github.com/sigstore/sigstore-conformance/pull/74/commits/79b2b1e721b310dfe2bc4f0dc1b5e4001cf2efc3)

test/client.py
tetsuo-cpp: [62ab292](https://github.com/sigstore/sigstore-conformance/pull/74/commits/62ab2927b6d17ab4a40113a3450c4fee97465fe3)

woodruffw (approved):
LGTM! :shipit:

PR #75, opened 2023-04-26
Fix our use of `GITHUB_TOKEN`

woodruffw:
This was my fault.


woodruffw:
This will fail because I broke it on `main`, but I tested on my fork and confirmed that `${{ github.token }}` works as expected.

Sorry for the churn here 🙂

bobcallaway (approved):

PR #93, opened 2023-07-31
Fix typo to reference `skip-signing` input; mark additional test as using signing

steiza:
#### Summary

This is a follow-on to https://github.com/sigstore/sigstore-conformance/issues/81, to allow testing verification libraries that don't support signing. There was a test missed that uses signing, and a typo in getting the input from the Action to the test driver.

#### Release Note

NONE

#### Documentation

N/A


woodruffw (approved):
Thanks!

PR #95, opened 2023-08-01
action, conftest: initial xfail support

### LINKED ISSUE #87: Add ability to feature-flag which tests are expected to fail
**Description**

As raised in the clients repo today, individual clients which run the conformance suite action should have a way to declare which tests they expect to fail due to not supporting some feature.

Something like this:

```yaml
- uses: sigstore/sigstore-conformance@v0.0.5
  with:
    entrypoint: sigstore
    xfail:
      - some_test_name
      - some_other_test_name
```

Tests that are listed in the action configuration should still be run, but should be `xfail`'d (i.e., if they're expected to fail and they don't fail, the test suite should fail)

This should probably also handle changes to test names between action versions, by failing if a test is configured to xfail and it doesn't exist.

### LINKED ISSUE #94: Add a flag for skipping bundle tests
We want to eventually make these mandatory, but in the interest of expanding conformance suite usage we should add a `skip-bundle` option similar to the `skip-signing` option that was added with https://github.com/sigstore/sigstore-conformance/issues/81.

woodruffw:
This adds an `xfail` input that configures a whitespace-separated list of test names to mark as `xfail`. These `xfail` tests are "strict", meaning that an unexpected pass with them causes a test failure, rather than a silent pass.

Some thoughts:

* Maybe it makes sense to support `fnmatch` syntax here, e.g. `test_signature_verify::*`?

Closes #87.

Closes #94.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -26,3 +26,5 @@ jobs:
         id: sigstore-conformance
         with:
           entrypoint: ${{ github.workspace }}/sigstore-python-conformance
+          xfail: |
+            test_verify_empty
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -26,5 +26,3 @@ jobs:
         id: sigstore-conformance
         with:
           entrypoint: ${{ github.workspace }}/sigstore-python-conformance
-          xfail: |
-            test_verify_empty
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,5 @@ dev: env/pyvenv.cfg
 .PHONY: lint
 lint: env/pyvenv.cfg $(ALL_PY_SRCS)
 	./env/bin/python -m black $(ALL_PY_SRCS)
-	./env/bin/python -m isort $(ALL_PY_SRCS)
-	./env/bin/python -m flake8 --max-line-length 100 $(ALL_PY_SRCS)
+	./env/bin/python -m ruff --fix $(ALL_PY_SRCS)
 	./env/bin/python -m mypy action.py test/
--- a/action.py
+++ b/action.py
@@ -28,17 +28,8 @@ def _debug(msg):
         print(f"\033[93mDEBUG: {msg}\033[0m", file=sys.stderr)
 
 
-def _log(msg):
-    print(msg, file=sys.stderr)
-
-
 def _sigstore_conformance(*args) -> int:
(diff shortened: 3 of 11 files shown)


woodruffw:
Example of strict xfail failing (since the tests actually pass):

```
============================= test session starts ==============================
platform linux -- Python 3.11.4, pytest-7.1.3, pluggy-1.2.0
rootdir: /home/runner/work/sigstore-conformance/sigstore-conformance
collected 15 items

test/test_bundle.py .......                                              [ 46%]
test/test_certificate_verify.py .                                        [ 53%]
test/test_signature_verify.py FF...                                      [ 86%]
test/test_simple.py ..                                                   [100%]

=================================== FAILURES ===================================
______________________ test_verify_empty[BundleMaterials] ______________________
[XPASS(strict)] skipped by suite runner
_______________ test_verify_empty[SignatureCertificateMaterials] _______________
[XPASS(strict)] skipped by suite runner
=========================== short test summary info ============================
FAILED test/test_signature_verify.py::test_verify_empty[BundleMaterials]
FAILED test/test_signature_verify.py::test_verify_empty[SignatureCertificateMaterials]
=================== 2 failed, 13 passed in 70.70s (0:01:10) ====================
```

woodruffw:
CC @steiza: I didn't change anything with this PR, but I _think_ we'll eventually want to replace the `skip-signing` option with this (or, perhaps, just rewrite it in terms of `xfail`).

steiza:
Yeah, on one hand we don't want to have to add a set of flags for every possible subset of functionality a client library supports, and supporting `xfail` is way easier than maintaining a set of flags.

On the other hand, when a client library updates to a new release of conformance testing, they'll have to manually go through any new tests that are failing, to determine if the failure indicates a bug or is using a feature their library doesn't support. Hopefully they are careful in their analysis!

`fnmatch` syntax is an interesting idea, especially for features that are contained in a single test file. But if we start to have features that overlap and try to name our functions to reflect that overlap... it would probably be easier to maintain flags.

After thinking through it, I'm okay with replacing `skip-signing` with `xfail` (or rewriting it to use `xfail`).

woodruffw:
Yeah, I think rewriting it to use `xfail` makes the most sense for now: signing is a "big" subset of functionality, so keeping a distinct option for it makes sense to me as a special case 🙂

woodruffw:
I'll do a follow-up with `fnmatch` support.

Issue #99, opened 2023-09-20
Make it easier to run verification tests locally


steiza:
**Description**

The existing instructions to run the tests locally involve using a GitHub token to get an OIDC JWT from https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon.

If you're only testing verification use-cases though, you don't need an OIDC JWT.

We should make it so if you run `pytest test --skip-signing ...` you don't need a GitHub token, since you don't need an OIDC JWT.

woodruffw:
Fully I agreed! I saw you opened #100; thank you!

PR #100, opened 2023-09-20
Make it easier to run verification test locally

steiza:
For #99

#### Summary

If you're only testing verification use-cases locally, you don't need the OIDC JWT.

You can test this change with:

```sh
(env) $ pytest test --skip-signing --entrypoint=SIGSTORE_CLIENT
```

#### Release Note

* Enabled easier local testing for verification with `pytest test --skip-signing --entrypoint=SIGSTORE_CLIENT`

#### Documentation

N/A

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -36,6 +36,10 @@ class OidcTokenError(Exception):
     pass
 
 
+class ConfigError(Exception):
+    pass
+
+
 def pytest_addoption(parser) -> None:
     """
     Add the `--entrypoint`, `--github-token`, and `--skip-signing` flags to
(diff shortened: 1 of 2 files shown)


test/conftest.py
```diff
@@ -52,7 +52,6 @@ def pytest_addoption(parser) -> None:
         "--github-token",
         action="store",
         help="the GitHub token to supply to the Sigstore client under test",
-        required=True,
```
steiza, line 55: This is maybe not idea - now if you leave off `--github-token` (and didn't specify `--skip-staging`) you'll get a bunch of `401 Client Error: Unauthorized for url: https://api.github.com/repos/sigstore-conformance/extremely-dangerous-public-oidc-beacon/actions/workflows/...` errors.

woodruffw: Hmm, perhaps we could test for `--skip-staging` and, if not present, make the absence of `--github-token` a hard error? That would handle the use case here while also simultaneously preventing a bunch of testcases that we know will fail 🙂

woodruffw (commented):
LGTM overall, one design thought!

test/conftest.py
steiza: I think https://github.com/sigstore/sigstore-conformance/pull/100/commits/3d8279cb3459a06076788a2c9e36ea0796b140eb should address this!

woodruffw (approved):

woodruffw:
Thanks @steiza!

PR #102, opened 2023-10-05
Allow multiple artifacts to exist

jku:
#### Summary

This relates to https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/issues/5

I'm planning to publish the (extremely-dangerous-public-oidc) token in GitHub Pages to make it easier to use. The only realistic way to do that is to use actions/upload-pages-artifact and then actions/deploy-pages -- the currently uploaded artifact is not usable for this.

This commit makes the script not fail if there are multiple artifacts, but still uses the same artifact as before. This makes it possible to start publishing a new 'github-pages' artifact without immediately breaking this script.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -137,18 +137,14 @@ def identity_token(pytestconfig) -> str:
     resp.raise_for_status()
 
     resp_json = resp.json()
-    artifacts = resp_json["artifacts"]
-    if len(artifacts) != 1:
-        raise OidcTokenError(
-            f"Found unexpected number of artifacts on OIDC beacon run: {artifacts}"
-        )
+    artifact_id = None
 
-    oidc_artifact = artifacts[0]
-    if oidc_artifact["name"] != "oidc-token":
-        raise OidcTokenError(
-            f"Found unexpected artifact on OIDC beacon run: {oidc_artifact['name']}"
-        )
-    artifact_id = oidc_artifact["id"]
+    for oidc_artifact in resp_json["artifacts"]:
+        if oidc_artifact["name"] == "oidc-token":
+            artifact_id = oidc_artifact["id"]
+
+    if not artifact_id:
+        raise OidcTokenError("Artifact 'oidc-token' could not be found")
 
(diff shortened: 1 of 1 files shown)


jku:
I have not actually run the tests locally yet, still figuring this out

jku:
I have failed to create a fine-grained token that would work for `--github-token` so I'd like to either:
* see the test results from the self test run in CI
* or get some advice on how to create a limited token for this

test/conftest.py
```diff
@@ -137,18 +137,14 @@ def identity_token(pytestconfig) -> str:
...
-        raise OidcTokenError(
-            f"Found unexpected artifact on OIDC beacon run: {oidc_artifact['name']}"
-        )
-    artifact_id = oidc_artifact["id"]
+    for oidc_artifact in resp_json["artifacts"]:
+        if oidc_artifact["name"] == "oidc-token":
+            artifact_id = oidc_artifact["id"]
+
+    if not artifact_id:
+        raise OidcTokenError("Artifact 'oidc-token' could not be found")
```
woodruffw, line 147: Nitpick, but maybe:

```suggestion
    try:
        artifact_id = next(a["id"] for a in resp_json["artifacts"] if a["name"] == "oidc-token")
    except StopIteration:
        raise OidcTokenError("Artifact 'oidc-token' could not be found")
```

woodruffw (approved):
LGTM. I ran the self-tests in CI and they look fine.

(One nitpick, but if you don't like it then we can ignore.)

test/conftest.py
jku: sure that's fine by me

jku:
ah looks like I missed a signed-off-by

woodruffw:
> ah looks like I missed a signed-off-by

If you're feeling lazy, I can override it in a moment 🙂 (since it all gets squashed down anyways)

woodruffw (approved):

PR #104, opened 2023-10-11
Use bundle v2 mediaType

loosebazooka:


loosebazooka:
actually I should just repurpose this as an issue

PR #111, opened 2023-12-05
README: prep 0.0.9

woodruffw:
Gets 0.0.9 ready.


steiza (approved):
Looks good to me!

This particular example will fail until https://github.com/sigstore/sigstore-python/issues/821 is fixed, but the example is meant to be illustrative, not copy-pasted, so I think it's fine.

haydentherapper (approved):

PR #115, opened 2023-12-15
Simplify DSSE bundle tests

bdehamer:
Updates the recently-added DSSE bundle tests to simplify the test set-up. Previously, the attestation subject was a binary artifact. With this update, the subject is now a simple text file (similar to the other test subjects).

I've also removed the RFC31616 timestamp from all of the test bundles -- due to an error in code used to generate the test bundles, the timestamp was invalid. When I've fixed the issue, I'll submit another PR to re-add test cases around RFC3161 timestamp verification.


woodruffw (approved):
LGTM! @steiza @tnytown could you give this a look as well?

steiza (approved):
Looks good to me!

PR #124, opened 2024-02-16
feat:checking that the client is allowed to run against staging

### LINKED ISSUE #121: Test against staging too
It eould be useful if this project also ensured that the client works with current staging infra:
* client developers would find out ASAP if their client breaks when something changes in staging: these changes are likely to propagate to production later
  * in case of infrastructure bug, client developer complaints could prevent the propagation of this bug
  * in case of client bug or missing feature this gives client developer more time to fix issues (or ask for delay in infrastructure change propagation)
  
There might be several ways to accomplish this but a simple one might be to  
* Add a `--staging` flag to the client-under-test CLI spec
* add a single new test `test_simple_staging()` that does exactly what `test_simple()` does but against staging

cc @loosebazooka

javanlacerda:
Closes #121 

#### Summary

One conformance requirement for the clients is that they should be able to run the commands against `staging`.  After this PR, the clients should be able to receive the `--staging` token in the command line and point to staging resources.

#### Release Note

Updated client run function to receive `staging` argument.
Updated run subprocess call to pass `--staging` flag to clients cli when required.
Modified CLI parser to receive --staging flag
Update CLI and conformance READMEs adding staging features

### DIFF
--- a/test/client.py
+++ b/test/client.py
@@ -111,7 +111,7 @@ class SigstoreClient:
     methods should not be called directly.
     """
 
-    def __init__(self, entrypoint: str, identity_token: str) -> None:
+    def __init__(self, entrypoint: str, identity_token: str, staging: bool) -> None:
         """
         Create a new `SigstoreClient`.
 
@@ -120,15 +120,20 @@ class SigstoreClient:
         self.entrypoint = entrypoint
         self.identity_token = identity_token
         self.completed_process: subprocess.CompletedProcess | None = None
+        self.staging = staging
 
     def run(self, *args) -> None:
         """
         Execute a command against the Sigstore client.
         """
         self.completed_process = None
+        full_command = [self.entrypoint, *args]
+        if self.staging:
+            full_command.append("--staging")
+
         try:
             self.completed_process = subprocess.run(
-                [self.entrypoint, *args],
+                full_command,
                 text=True,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT,
@@ -182,8 +187,7 @@ class SigstoreClient:
         This is an overload of `sign` for the signature/certificate flow and should not
         be called directly.
         """
-
-        self.run(
+        args = [
             "sign",
             "--identity-token",
             self.identity_token,
@@ -192,7 +196,9 @@ class SigstoreClient:
             "--certificate",
             materials.certificate,
             artifact,
-        )
+        ]
+
+        self.run(*args)
 
     @sign.register
     def _sign_for_bundle(self, materials: BundleMaterials, artifact: os.PathLike) -> None:
@@ -202,21 +208,19 @@ class SigstoreClient:
         This is an overload of `sign` for the bundle flow and should not be called directly.
         """
 
-        self.run(
+        args = [
             "sign-bundle",
             "--identity-token",
             self.identity_token,
             "--bundle",
             materials.bundle,
             artifact,
-        )
+        ]
+
+        self.run(*args)
 
     @singledispatchmethod
-    def verify(
-        self,
-        materials: VerificationMaterials,
-        artifact: os.PathLike,
-    ) -> None:
+    def verify(self, materials: VerificationMaterials, artifact: os.PathLike) -> None:
         """
         Verify an artifact with the Sigstore client. Dispatches to `_verify_for_sigcrt`
         when given `SignatureCertificateMaterials`, or `_verify_for_bundle` when given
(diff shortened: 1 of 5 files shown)


test/client.py
```diff
@@ -159,8 +159,13 @@ def raises(self):
             )
             raise ClientUnexpectedSuccess(msg)
 
+    def set_staging(self, args):
+        args.append("--staging")
```
woodruffw, line 163: Nitpick: this helper is tiny, so I think we can just inline it 🙂

woodruffw, line 163:
```suggestion
```

test/client.py
```diff
@@ -192,30 +196,39 @@ def _sign_for_sigcrt(
             "--certificate",
             materials.certificate,
             artifact,
-        )
+        ]
+
+        if staging:
+            self.set_staging(args)
```
woodruffw, line 202:
```suggestion
            args.append("--staging")
```

test/client.py
```diff
@@ -192,30 +196,39 @@ def _sign_for_sigcrt(
...
             "--identity-token",
             self.identity_token,
             "--bundle",
             materials.bundle,
             artifact,
-        )
+        ]
+
+        if staging:
+            self.set_staging(args)
```
woodruffw, line 225:
```suggestion
            args.append("--staging")
```

test/test_simple_staging.py
```diff
@@ -0,0 +1,24 @@
+from test.conftest import _MakeMaterials
+
+import pytest  # type: ignore
+
+from .client import SigstoreClient
+
+
+@pytest.mark.signing_staging
```
woodruffw, line 8: I might be missing it, but where does this `signing_staging` marker come from?

woodruffw (commented):
LGTM overall, but with one broader design thought: rather than enabling individual tests to be run against staging, maybe we should allow the *entire suite* to be flipped between production and staging at a time? 

My rationale: at least for testcases that do end-to-end signing (meaning they aren't implicitly tied to a specific instance already), we expect them to work on both instances and probably don't want to duplicate the code for both. If the entire suite supports being run in either mode (with skips where necessary for tests that are hard-baked against one or the other), then we get additional coverage against both instances for free.

Thoughts? This may be more effort than is immediately worth 🙂

test/client.py
```diff
@@ -282,4 +298,7 @@ def _verify_for_bundle(self, materials: BundleMaterials, artifact: os.PathLike)
         if getattr(materials, "trusted_root", None) is not None:
             args.extend(["--trusted-root", materials.trusted_root])
 
+        if staging:
+            args.append("--staging")
```
woodruffw, line 302: I think we also need to add `--staging` to the CLI protocol for the conformance suite, here: https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md

test/test_simple_staging.py
javanlacerda: It was wrong!! But I already removed as I implemented to run the tests against staging by the pytest command instead of create another test.

jku:
> rather than enabling individual tests to be run against staging, maybe we should allow the entire suite to be flipped between production and staging at a time?

We can try that... but it may make sense to still enable just a subset of tests (the end-to-end ones) in the beginning: otherwise it might be quite a bit of work to re-create all of the test assets for staging (I assume that will be needed at least: I haven't created any of the assets)

If all tests can be run against staging, then we need to provide clients a way to enable/disable staging testing: xfail for individual tests is not viable... Some potential solutions:
* we add some sort of suite wide xfail value for all-staging-tests or 
*  we add a `staging` option to the GH action: this way clients can add a whole new action call to to their workflow when they are ready to test staging

I suppose the second one makes more sense?

jku:
Waving hands about what tests should be executed:
* --skip-signing already uses  `@pytest.mark.signing` to skip tests: we could do something similar: if pytest gets "--staging", it could skip tests with `@pytest.mark.no_staging_assets`
* when/if we actually end up adding some staging assets, we need to figure out how to load the correct ones: I'm sure we can modify the `make_materials*()` functions at that time

jku:
Some of results that currently PASS when running `GHA_SIGSTORE_CONFORMANCE_XFAIL="test_verify_with_trust_root test_verify_dsse_bundle_with_trust_root" pytest test --staging --skip-signing   --entrypoint=$PWD/sigstore-python-conformance` are likely incorrect: if the suite uses production assets against staging infra, the client-under test is going to fail (but it's failing for the wrong reasons) .

woodruffw:
Yeah, getting the production/staging states right will probably require a larger refactor/discussion of how we want the CI to run the suite. I'm okay with punting on that for now 🙂

woodruffw (approved):
LGTM, thank you @javanlacerda!

woodruffw:
(CCing @jku for review as well, since I'm now the last pusher.)

di (approved):

di:
Ah, sorry @jku I merged too quickly here, please feel free to finish your review and suggest any changes if we missed something!

haydentherapper:
Should we expect a significant increase in traffic to staging?

PR #125, opened 2024-02-20
Improve dev experience

### LINKED ISSUE #123: developer experience could be better
It's kind of hard to test out conformance or to start developing it further, especially without a bunch of Python experience. I think we should make it easier to get from "git clone" to "first successful test suite run"

Some specific points of friction:
* virtual env in Makefile is not mentioned in README
* It's not clear that I need to install sigstore-python to test the test suite locally.
* is there a reason for `--entrypoint` to require an absolute path?
* figuring out the correct xfail incantation on CLI requires quite a bit of investigation

Potential quick fixes:
* Mention `make dev` and `source env/bin/activate` in Development section of README
* add sigstore into dev-requirements.txt: this would mean workflows/ci.yml ends up installing sigstore without needing it but maybe that's fine
* Add an example call like this into the README:
    ```
    GHA_SIGSTORE_CONFORMANCE_XFAIL="test_verify_with_trust_root test_verify_dsse_bundle_with_trust_root" \
        pytest test --skip-signing --entrypoint=$PWD/sigstore-python-conformance

    ```
 
 cc @javanlacerda

jku:
* Allow relative paths in --entrypoint
* Expand README with examples that run out-of-the-box
* Install sigstore in dev-requirements

Fixes #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,25 +59,41 @@ invoke the client.
 
 ## Development
 
-Running the conformance suite locally,
-
+Easiest way to run the conformance suite locally is with the provided virtual environment:
 ```sh
-(env) $ pytest test --entrypoint=SIGSTORE_CLIENT --identity-token=$(gh auth token)
+$ make dev
+$ source env/bin/activate
+(env) $
 ```
 
-Or if you are only checking verification use cases,
+The test suite can be configured with
+* `--entrypoint=$SIGSTORE_CLIENT` where SIGSTORE_CLIENT is path to a script that implements the
+  [CLI specification](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md)
+* `--identity-token=$GITHUB_TOKEN` where GITHUB_TOKEN is a GitHub token with actions:read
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -59,25 +59,41 @@ invoke the client.
...
+(env) $
 ```
 
-Or if you are only checking verification use cases,
+The test suite can be configured with
+* `--entrypoint=$SIGSTORE_CLIENT` where SIGSTORE_CLIENT is path to a script that implements the
+  [CLI specification](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md)
+* `--identity-token=$GITHUB_TOKEN` where GITHUB_TOKEN is a GitHub token with actions:read
+  access for public repositories (--identity-token is only required for signing tests)
+* optional (and currently experimental) `--staging`: This instructs the test suite to run
```
jku, line 74: I've marked `--staging` as experimental as I don't think it works correctly yet

woodruffw (approved):
LGTM, thank you @jku! Agreed on documenting `--staging` as experimental for now.

PR #127, opened 2024-02-23
Add staging to selftest

### LINKED ISSUE #121: Test against staging too

jku:
This might be enough to close #121.

#### Summary

Do several small changes required to support staging tests in self-test
* the `enable-staging` action input defaults to false: we may want to reconsider before releasing -- maybe opt-out is better?
* only a few tests are marked as compatible with staging at this point: I may have missed some but since there are loads of false negatives in the remaining tests (tests that seemingly have the correct result but only because they fail for the wrong reasons on staging) I didn't want to extend without research

#### TODO?

* `--staging` flag is sort of  documented in cli_protocol.md but it could probably be improved 

Test run https://github.com/sigstore/sigstore-conformance/actions/runs/8017481217

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -25,4 +25,5 @@ jobs:
         id: sigstore-conformance
         with:
           entrypoint: ${{ github.workspace }}/sigstore-python-conformance
+          enable-staging: true
           xfail: "test_verify_with_trust_root test_verify_dsse_bundle_with_trust_root"
--- a/README.md
+++ b/README.md
@@ -34,28 +34,34 @@ Some general testing principles for this suite are:
 
 ## Usage
 
-Simply add `sigstore/sigstore-conformance` to one of your workflows:
+1. Include an executable in your project that implements the
+client-under-test [CLI protocol](docs/cli_protocol.md).
+2. Use the `sigstore/sigstore-conformance` action in your test workflow:
+    ```yaml
+    jobs:
+      conformance:
+        runs-on: ubuntu-latest
+        steps:
+          - uses: actions/checkout@v4
+          # insert your client installation steps here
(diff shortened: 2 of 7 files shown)


jku:
cc @javanlacerda please have a look: did I miss anything?

javanlacerda:
> cc @javanlacerda please have a look:
> 
> * did I miss anything?
> * does the approach of doing two runs of the test suite in a single GH action call make sense?

Hey @jku! It seems awesome!

I am just wondering if we could have staging as default and production as optional. I mean, It could decrease the production infrastructure usage for development proposes. Does it make sense?

javanlacerda (approved):
LGTM! Thank you!

jku:
> I am just wondering if we could have staging as default and production as optional. I mean, It could decrease the production infrastructure usage for development proposes. Does it make sense?

I think the default should be to test everything (and not testing with production seems like a decision that could bite a client developer in the behind). Whether there is a use case to ever test only staging I don't know...

woodruffw:
> Whether there is a use case to ever test only staging I don't know...

I personally can't think of one 🙂 -- IMO we can save ourselves a state here and make staging tests "additive" rather than dealing with different combinations.

PR #131, opened 2024-03-14
Tweak staging invocation

### LINKED ISSUE #130: how should we enable staging
So I'm looking at doing some final testing before enabling staging test by default (changing the default value of the action input) however...

I started thinking about how this goes in the future (assume we have many more tests that support testing on staging):
* do we need to enable xfail for staging but not for prod? Probably yes because both get run on the same action invocation
* we could invent a syntax of our own (`xfail: "test_name@staging"` or something) and handle that in `conformance_xfail()` fixture -- or add another input to the action like `xfail_staging`

Alternatively we might want to require running the action separately for staging: this makes it easy to manage the xfail lists separately.

jku:
Main change here is the staging invocation: I'm proposing that a single action invocation only runs the testsuite once because 
* xfails becames painful if staging expected failures are not the same as prod (which seems like a normal thing at times).
* The test results are much clearer ("Staging conformance failure" is not as critical as "production conformance failure")

Other changes in this PR (from testing and reviewing for a hopefully near future release):
* run selftest daily (and on dispatch)
* update linter config to silence warnings
* update cli protocol doc and move `--staging` so it's not after the artifact filename

Fixes #130

I'll mark William a reviewer since we originally agreed to run test suite twice but  I think we may have been wrong.

--- 

For release Notes:

* The action now supports testing against Sigstore staging infrastructure: if your Sigstore client can be used with staging, please enable this by adding a second step to your workflow: see [README](https://github.com/sigstore/sigstore-conformance?tab=readme-ov-file#usage)

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -6,6 +6,8 @@ on:
       - main
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron: '10 10 * * *'
 
 jobs:
   selftest:
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
 
 jobs:
   selftest:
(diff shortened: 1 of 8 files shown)


woodruffw:
> I'm proposing that a single action invocation only runs the testsuite once because
> 
> * xfails becames painful if staging expected failures are not the same as prod (which seems like a normal thing at times).
> * The test results are much clearer ("Staging conformance failure" is not as critical as "production conformance failure")

Makes sense to me, thanks for documenting!

.github/workflows/conformance.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: '10 10 * * *'
```
woodruffw, line 10: Thanks for adding these!

woodruffw (approved):

PR #132, opened 2024-03-14
requirements: bump sigstore-protobuf-specs

woodruffw:
Bumps sigstore-protobuf-specs to `~= 0.3.0`, which includes v3 bundle support. Without this bump, anything in sigstore-python that constructs v3 bundles under the hood will fail (hooray for hidden version conflicts in Python environments!)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pytest==7.1.3
 requests==2.31.0
 cryptography>=39
-sigstore-protobuf-specs~=0.2.0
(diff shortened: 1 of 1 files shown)


woodruffw:
Example breakage here: https://github.com/sigstore/sigstore-python/pull/937

jku (approved):
Seems reasonable.

There is likely a way to keep the two environments separate (one for testrunner and a completely separate one for client-under-test) but this seems harmless and easier.

PR #134, opened 2024-03-16
Don't require a GitHub token

jku:
The new extremely-dangerous-public-oidc-beacon `current-token` branch allows us to fetch the JWT without a GitHub token. This makes local tests much nicer.

Changes:
* git clone the beacon repo to get the token
* Do this in a loop until token is valid
* Cache the token so we only do this once in a test run
* Remove all mention of github token in arguments, comments and docs

Some additional notes:
* retry times are tweaked from the values upstream (and special values are set for the interactive case)
* The caching could be smarter (could return cached value only if it's valid) but maybe this simple thing works?

#### Release Note

* `--github-token` is no longer required to run signing tests


jku:
This makes the test run significantly faster as well.

That is just a result of the caching which we could have done with the previous approach as well so it's kind of unrelated.

woodruffw (approved):
Awesome, thanks @jku!

PR #135, opened 2024-03-19
Capture stdout & stderr separately

segiddins:
#### Summary

I was running the conformance tests, and having stdout & stderr shown separately made debugging easier.

I figure this is what was intended since the error already has separate sections for stdout & stderr

#### Release Note

NONE

### DIFF
--- a/test/client.py
+++ b/test/client.py
@@ -134,7 +134,7 @@ class SigstoreClient:
                 full_command,
                 text=True,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
+                stderr=subprocess.PIPE,
                 check=True,
             )
         except subprocess.CalledProcessError as cpe:


woodruffw:
Yep, that looks right to me. Thanks for the fix!

woodruffw (dismissed):

test/client.py
```diff
@@ -134,7 +134,7 @@ def run(self, *args) -> None:
                 full_command,
                 text=True,
                 stdout=subprocess.PIPE,
```
woodruffw, line 136:
```suggestion
```

test/client.py
```diff
@@ -134,7 +134,7 @@ def run(self, *args) -> None:
...
-                stderr=subprocess.STDOUT,
+                stderr=subprocess.PIPE,
```
woodruffw, line 137:
```suggestion
                capture_output=True,
```

woodruffw (approved):

jku (approved):

PR #139, opened 2024-03-20
Add failure cases for checkpoints

loosebazooka:
a.txt.checkpoint_bad_keyhint.sigstore
modify the first base64 character of the signature, to affect the keyhint for that signature line

a.txt.checkpoint_invalid_signature.sigstore
modify the signature so it no longer validates

a.txt.checkpoint_wrong_roothash.txt
replace the checkpoint with an otherwise valid checkpoint from another bundle (from the same log instance)

This should catch any client *not* verifying a checkpoint from a bundle. (like sigstore-java was)


loosebazooka:
@william since we were talking about this on slack.

woodruffw:
Thanks @loosebazooka! Looking today. FYI @william is not me 😉

loosebazooka:
loool, poor @william I'm sorry. I don't why auto-fill did this to them.

woodruffw (approved):
Thanks @loosebazooka, looks great!

PR #142, opened 2024-05-07
Use .sigstore.json as bundle extension

loosebazooka:
Just a rename to match the spec.


woodruffw (approved):
Thanks @loosebazooka!

PR #143, opened 2024-05-08
Isolate staging/production self tests from eachother

loosebazooka:
should help determine quickly if failures are isolated to staging flakiness


woodruffw (approved):
Thanks!

PR #147, opened 2024-05-21
Fix lint issues

jku:
* Avoid variable reuse when variable type is different: this makes mypy happier
* include ruff automatic fixes


jku:
blocks #146

woodruffw (approved):

PR #152, opened 2024-08-16
dev-requirements: switch to sigstore-python 3.x

woodruffw:
This has been out for a while, let's see if this breaks any of the automation.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -25,7 +25,7 @@ jobs:
           cache: "pip"
 
       - name: install sigstore-python
-        run: pip install "sigstore ~= 2.0"
+        run: pip install "sigstore ~= 3.0"
 
       - name: conformance test sigstore-python
         uses: ./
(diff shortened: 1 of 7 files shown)


.github/workflows/conformance.yml
```diff
@@ -25,11 +25,11 @@ jobs:
...
-        run: pip install "sigstore ~= 2.0"
+        run: pip install "sigstore ~= 3.0"
 
       - name: conformance test sigstore-python
         uses: ./
         with:
           entrypoint: ${{ github.workspace }}/sigstore-python-conformance
           environment: ${{ matrix.sigstore-env }}
-          xfail: "test_verify_with_trust_root test_verify_dsse_bundle_with_trust_root test_verify_v_0_3"
+          xfail: "test_verify_with_trust_root test_verify_dsse_bundle_with_trust_root"
```
woodruffw, line 35: This no longer xfails, since 3.x supports v0.3 bundles.

(Verifying a DSSE bundle should also work though, so I need to look into that one.)

woodruffw: Ah, the problem there is that the "DSSE" bundle is actually an "intoto" entry, not a DSSE entry. I'll rename the test in a follow-up.

segiddins (approved):

jku (approved):
sounds reasonable

di (approved):

Issue #153, opened 2024-08-19
"DSSE" bundle tests are really intoto bundle tests


woodruffw:
Noticed this during #152: the `*dsse*` labeled tests are really tests of `intoto` entries (that use DSSE for the message part, but not the log entry part). As part of improving testcase categorization, we should probably adopt a scheme like this:

```
testcase_tlogkind_messagekind.bundleversion.good|bad.sigstore.json
```

so, for example, `d.txt.good.sigstore.json` would become:

```
d_intoto_dsse.v0.2.good.sigstore.json
```

(Where `d` should also really be a better name, but baby steps 🙂)

haydentherapper:
@segiddins also pointed this out to me when reviewing the Rekor v2 doc. We should add some tests for dsse entries as well.

loosebazooka:
> @segiddins also pointed this out to me when reviewing the Rekor v2 doc. We should add some tests for dsse entries as well.

Adding some now

haydentherapper:
Good to close?

jku:
> Good to close?

not really -- it's really misleading that the happy path dsse tests are dsse but the actually interesting (negative) tests are actually intoto: at least sigstore-python just fails to handle the kindversion so seemingly passes the tests... but has not actually been tested at all.

* should do some renaming of tests at a minimum
* preferably we would have the interesting tests implemented with the commonly supported kindversions first
* similar tests should be constructed for rekor v2 types

loosebazooka:
this appears to be fixed

PR #155, opened 2024-08-19
test: rename some assets

woodruffw:
Per #153: this is nowhere near complete, but makes the relationships between a few assets and their filenames slightly clearer.


test/test_bundle.py
```diff
@@ -8,39 +8,39 @@
...
+def test_verify_intoto_dsse_bundle_with_trust_root(
     client: SigstoreClient, make_materials_by_type: _MakeMaterialsByType
 ) -> None:
     """
     Test the happy path of verification for DSSE bundle w/ custom trust root
     """
     materials: BundleMaterials
     input_path, materials = make_materials_by_type("d.txt", BundleMaterials)
-    materials.bundle = Path("d.txt.good.sigstore.json")
+    materials.bundle = Path("d.txt+custom-trust-root_intoto_dsse.good.v0.2.sigstore.json")
```
loosebazooka, line 43: is the use of `+` intentional?

woodruffw: Yeah, that was me trying to think of a reasonable delimiter to split the "original signed filename" component from the "explanation of what this tests" component. We could pick another delimiter, though, if you think that one's confusing!

loosebazooka: yeah seems like we're putting the target infra into this filename when our naming schema doesn't support it? Not sure if another parameter on the filename for staging/prod/custom-trust-root is the right option though (especially since we configure lots of staging tests via flag).

Maybe naming for this one case doesn't matter that much yet. We can rename later :shrug: ?

woodruffw: Yeah, I'm starting to think that there's no good naming convention that'll encapsulate all of the varieties of states we want to test. I'm going to close this for now while I think about alternative approaches.

PR #156, opened 2024-09-13
Conformance tests for CPython release signatures

woodruffw:
~~WIP.~~

This adds conformance suite tests for CPython's releases, which use Sigstore bundles. The idea is to continuously check whether CPython releases are verifiable by multiple clients.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -27,6 +27,13 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Check out latest CPython release metadata
+      uses: actions/checkout@v4
+      with:
+        repository: woodruffw/cpython-release-tracker
+        persist-credentials: false
+        path: cpython-release-tracker
+
     - name: Set up sigstore-conformance
       run: |
         # NOTE: Sourced, not executed as a script.
--- a/action.yml
+++ b/action.yml
@@ -33,6 +33,7 @@ runs:
         repository: woodruffw/cpython-release-tracker
         persist-credentials: false
         path: cpython-release-tracker
+        ref: 31c7721f29ae4181484ae9b399ecc5f04da947e1
 
     - name: Set up sigstore-conformance
       run: |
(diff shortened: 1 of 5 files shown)


woodruffw:
This is good to go, except for two subtest failures: it looks like `python3101.chm` and `python31011.chm` both still have old (invalid) bundles.

di (approved):

action.yml
```diff
@@ -27,6 +27,13 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Check out latest CPython release metadata
+      uses: actions/checkout@v4
```
segiddins, line 31: Should this pin to a SHA for the checkout action, and the repo being checked out?

segiddins (approved):

action.yml
woodruffw: Yeah, good call. I'll ping the ref for the checkout itself but I'll defer pinning the action to a follow-up PR (since a bunch should probably be done all at once).

jku (approved):
Looks good -- I wish the external repo wasn't needed but that just shows the missing links in this chain and isn't a problem of this PR

woodruffw:
> Looks good -- I wish the external repo wasn't needed but that just shows the missing links in this chain and isn't a problem of this PR

Yeah, agreed -- if people would prefer, I'd happily have this org adopt the tracking repo.

test/test_bundle.py
```diff
@@ -351,3 +355,65 @@ def test_verify_rejects_checkpoint_with_no_matching_key(
 
     with client.raises():
         verify_bundle(materials, input_path)
+
+
+def test_verify_cpython_release_bundles(subtests, client):
+    cpython_release_dir = Path(os.getenv("GITHUB_WORKSPACE")) / "cpython-release-tracker"
```
segiddins, line 361: this causes an exception when `GITHUB_WORKSPACE` is not set

Issue #157, opened 2024-09-13
Conformance suite: support verifying by hash instead of file


woodruffw:
This is a blocker for #156: we don't want to have to download every CPython release artifact to verify it, especially since we already have their hashes.

To accomplish this, the `verify-bundle` and `verify` commands will probably need to learn to take digests as well, e.g.

```bash
${ENTRYPOINT} verify ... sha256:hexstringhere
```

This shouldn't be too difficult to plumb in, although clients will need to learn about it.

Core parts here:

1. Document hashes as an acceptable input in the CLI protocol: <https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md>
2. Plumb them into the sigstore-python conformance suite helper: <https://github.com/sigstore/sigstore-conformance/blob/main/sigstore-python-conformance>
    * This may require us to move away from a "basic" wrapper over the `sigstore` CLI and towards using the `sigstore` APIs directly.

CC @facutuesca

segiddins:
If we're refactoring the conformance interface, I think it makes sense to have the input be the canonical proto for verification? https://github.com/sigstore/protobuf-specs/blob/b288f87c935c02102866e2c8df2f3d66a812f446/protos/sigstore_verification.proto#L144

steiza:
sigstore-go supports verifying an artifact by providing the digest instead of the artifact itself - this is also quite useful when verifying large container images.

I haven't looked closely at the verification protobuf, but does it support supplying the artifact digest? I didn't see it when giving it a quick look over.

loosebazooka:
I don't know if anyone is using the verification proto directly yet. But it does want an actual artifact reference or actual artfiact: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_verification.proto#L133 -- we probably have to edit it to allow digests.

sigstore-java's API takes either a digest or a file reference. But our CLI doesn't care because it's not published -- happy to integrate with whatever.

woodruffw:
(@segiddins and I talked about this IRL, so summarizing here).

> If we're refactoring the conformance interface, I think it makes sense to have the input be the canonical proto for verification? [sigstore/protobuf-specs@`b288f87`/protos/sigstore_verification.proto#L144](https://github.com/sigstore/protobuf-specs/blob/b288f87c935c02102866e2c8df2f3d66a812f446/protos/sigstore_verification.proto#L144)

[...]

PR #158, opened 2024-09-18
Support verifying digests in addition to artifacts

facutuesca:
This PR adds support for verifying hashes (instead of files) to the CLI protocol (only when verifying bundles).

 It also adapts the `sigstore-python-conformance` helper to support it, given `sigstore-python` supports verifying hashes [since v3.3.0](https://github.com/sigstore/sigstore-python/releases/tag/v3.3.0).

All the existing bundle verification tests have been parametrized to also test the hash verification, in addition to the existing `Path` (file) verification.

This closes https://github.com/sigstore/sigstore-conformance/issues/157

cc @woodruffw

### DIFF
--- a/docs/cli_protocol.md
+++ b/docs/cli_protocol.md
@@ -77,7 +77,7 @@ ${ENTRYPOINT} verify [--staging] --signature FILE --certificate FILE --certifica
 #### Bundle flow
 
 ```console
-${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] FILE
+${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] [--verify-digest] FILE_OR_DIGEST
 ```
 
 | Option | Description |
@@ -87,4 +87,5 @@ ${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDE
 | `--certificate-identity IDENTITY` | The expected identity in the signing certificate's SAN extension |
 | `--certificate-oidc-issuer URL` | The expected OIDC issuer for the signing certificate |
 | `--trusted-root` | The path of the custom trusted root to use to verify the bundle |
-| `FILE` | The path to the artifact to verify |
+| `--verify-digest` | Presence indicates client should interpret `FILE_OR_DIGEST` as a digest. |
+| `FILE_OR_DIGEST` | The path to the artifact to verify, or its digest. The digest should start with the `sha256:` prefix. |
--- a/test/test_bundle.py
+++ b/test/test_bundle.py
@@ -5,10 +5,14 @@ from cryptography import x509
 from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
 
 from test.client import BundleMaterials, SigstoreClient
-from test.conftest import _MakeMaterialsByType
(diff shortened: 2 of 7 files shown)


test/test_bundle.py
```diff
@@ -8,56 +8,64 @@
 from test.conftest import _MakeMaterialsByType
 
 
-def test_verify(client: SigstoreClient, make_materials_by_type: _MakeMaterialsByType) -> None:
+@pytest.mark.parametrize("use_digest", [True, False])
```
woodruffw, line 11: This PR looks great overall, although I wonder if there's a cleaner way to implicitly parametrize all of these tests -- the current explicit parametrization IMO makes them a little harder to read, and adds a bit of stock boilerplate throughout.

(If not, I'm fine with this -- but we should confirm that there's no less verbose alternative that we're missing 🙂)

facutuesca: Yeah I had the same feeling, I'll take a look to see if I can find a cleaner way to do it

docs/cli_protocol.md
```diff
@@ -87,4 +87,5 @@ ${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDE
 | `--certificate-identity IDENTITY` | The expected identity in the signing certificate's SAN extension |
 | `--certificate-oidc-issuer URL` | The expected OIDC issuer for the signing certificate |
 | `--trusted-root` | The path of the custom trusted root to use to verify the bundle |
-| `FILE` | The path to the artifact to verify |
+| `--verify-digest` | Presence indicates client should interpret `FILE_OR_DIGEST` as a digest. |
```
segiddins, line 90: it looks like this should move up a line

test/test_bundle.py
segiddins: I think it might be possible by overriding `make_materials_by_type`?

facutuesca: I tried parametrizing `make_materials_by_type`, the problem is that it's also used to generate materials for signing, so overriding it (or parametrizing it) to return a digest instead of an `Path` makes things more complicated.

I ended up creating a fixture for verifying bundles (`verify_bundle`), which is parametrized to run twice: once with the artifact's Path, and then with the artifact's digest. This removes the parametrization from the unit tests themselves, at the cost of adding an extra fixture to the parameter list.

docs/cli_protocol.md
facutuesca: Do you mean that `--verify-digest` should be above `--trusted-root` ?

woodruffw: I think the diff is just confusing here -- the current location looks fine to me, since it's an optional flag like `--trusted-root`.

woodruffw (approved):
LGTM, nice work @facutuesca!

This is good to go, but CCing @steiza @kommendorkapten @loosebazooka as other client maintainers who will probably want to be aware of this as it goes in -- you'll likely need to update your conformance wrappers!

steiza:
Adding @bdehamer for sigstore-js.

This should be fine for sigstore-go - it's easy for us to set our verification policy accordingly. I don't think this will work with cosign, which maps `verify-bundle` to `cosign verify-blob`, which only supports URLs and files, not digests.

But that's not a problem with this pull request - we'll have to figure out how to have cosign additionally handle digests, maybe as part of the `VerifierOptions` @codysoyland suggested in https://github.com/sigstore/cosign/issues/3879.

segiddins (approved):

woodruffw:
Thanks @facutuesca, nice work here!

PR #160, opened 2024-10-02
action: bump cpython-release-tracker ref

woodruffw:
For the latest releases: https://github.com/woodruffw/cpython-release-tracker/pull/5

I need to figure out a good way to automate this PR as well.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -33,7 +33,7 @@ runs:
         repository: woodruffw/cpython-release-tracker
         persist-credentials: false
         path: cpython-release-tracker
-        ref: 31c7721f29ae4181484ae9b399ecc5f04da947e1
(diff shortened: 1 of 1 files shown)


jku:
I think dependabot supports git submodules at some level maybe that would work?



(I cannot believe I'm suggesting using submodules for anything)

woodruffw:
> I think dependabot supports git submodules at some level maybe that would work?
> 
> (I cannot believe I'm suggesting using submodules for anything)

Yeah, I could do this with a submodule but then we're using submodules :lolsob:



> would this work with a cron? just reading the latest commit off of that repo?

I think so -- PyCA Cryptography has a similar "update this ref in the workflow with a cron workflow" that I can crib.

PR #162, opened 2024-10-02
Remove --verify-digest flag

### LINKED ISSUE #161: Remove --verify-digest flag, clients should parse the target
https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md#bundle-flow-1 has `--verify-digest` flag to select how to handle the positional argument `FILE_OR_DIGEST`.

Clients should just consider anything that starts with `sha256:` as a digest and everything else as a file. We can remove the `--verify-digest` flag.

loosebazooka:
fixes #161, this doesn't remove the feature developed in #158, only removes the `--verify-digest` flag instead making clients parse the argument to determine in the action.

### DIFF
--- a/sigstore-python-conformance
+++ b/sigstore-python-conformance
@@ -15,8 +15,6 @@ SUBCMD_REPLACEMENTS = {
 ARG_REPLACEMENTS = {
     "--certificate-identity": "--cert-identity",
     "--certificate-oidc-issuer": "--cert-oidc-issuer",
-    # sigstore-python detects if the input is a file path or a digest without needing a flag
-    "--verify-digest": None,
 }
 
 # Trim the script name.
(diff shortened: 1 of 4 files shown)


sigstore-python-conformance
```diff
@@ -15,8 +15,6 @@ SUBCMD_REPLACEMENTS = {
 ARG_REPLACEMENTS = {
     "--certificate-identity": "--cert-identity",
     "--certificate-oidc-issuer": "--cert-oidc-issuer",
-    # sigstore-python detects if the input is a file path or a digest without needing a flag
-    "--verify-digest": None,
```
woodruffw, line 19: There's a line below where we remove `None` from the replacement list, could you remove that as well? I think it's no longer necessary if we're removing this.

woodruffw (approved):
Thanks @loosebazooka! One small nitpick.

sigstore-python-conformance
loosebazooka: gotcha, yeah

PR #163, opened 2024-10-02
cli_protocol: clarify FILE_OR_DIGEST

woodruffw:
@phillmv pointed out that the current conformance definition is a little ambiguous, and can be tripped up on clients like `gh` (which emits `sha256:hash.jsonl` files by default). This clarifies it to say that anything that isn't *exactly* a hash should be intepreted as a filepath.


loosebazooka (approved):

loosebazooka:
what's the precedence? filename or hash first?

PR #164, opened 2024-10-16
Bump cpython artifacts

woodruffw:
See https://github.com/woodruffw/cpython-release-tracker/pull/7.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -33,7 +33,7 @@ runs:
         repository: woodruffw/cpython-release-tracker
         persist-credentials: false
         path: cpython-release-tracker
-        ref: 331017869cd86699aaae8f4fadf003e4e36b1c2e
+        ref: 124a4005d3822cc21f16396e81bf08af3e858561
 
     - name: Set up sigstore-conformance
       run: |
--- a/action.yml
+++ b/action.yml
@@ -33,7 +33,7 @@ runs:
         repository: woodruffw/cpython-release-tracker
         persist-credentials: false
         path: cpython-release-tracker
-        ref: 9eb74dae56e55433ee03c7283dc013645d5d8c4e
+        ref: 331017869cd86699aaae8f4fadf003e4e36b1c2e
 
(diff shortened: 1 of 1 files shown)


kommendorkapten:
The change looks ok in that it references the merge commit, give me a ping when the issues with verification are sorted out 👍

jku:
`Certificate's OIDCIssuer does not match (got 'https://github.com/login/oauth', expected 'https://accounts.google.com')`

does the cpython release system allow any kind of signature or how does this happen?

di:
> does the cpython release system allow any kind of signature or how does this happen?

It allows any issuer, and this looks like it's due to https://github.com/sigstore/sigstore-python/issues/1204

woodruffw:
Yep, that was the problem -- sorry, I forgot to follow up on this earlier. Now that that's fixed I should be able to rebuild that table and this will pass.

di:
I think you still need https://github.com/woodruffw/cpython-release-tracker/pull/8 as well, no?

woodruffw:
> I think you still need [woodruffw/cpython-release-tracker#8](https://github.com/woodruffw/cpython-release-tracker/pull/8) as well, no?

Yep! I was going to re-run the script to pull that change in, but a manual change to it also works 🙂

action.yml
```diff
@@ -33,7 +33,7 @@ runs:
         repository: woodruffw/cpython-release-tracker
         persist-credentials: false
         path: cpython-release-tracker
-        ref: 9eb74dae56e55433ee03c7283dc013645d5d8c4e
+        ref: 331017869cd86699aaae8f4fadf003e4e36b1c2e
```
woodruffw, line 36:
```suggestion
        ref: 124a4005d3822cc21f16396e81bf08af3e858561
```

woodruffw:
Yep! Looks like we're green again here, so CCing @kommendorkapten et al for approval 🙂

haydentherapper (approved):

PR #165, opened 2024-11-21
Create provenance-generator

loosebazooka:
Create provenance for a.txt from within this repo

This is workflow dispatch only. It will generate official github provenance which can be copied into the test/assets directory for use with dsse-intoto bundle tests.


loosebazooka:
this is maybe pointless.

PR #166, opened 2024-11-21
Test fail on mismatched hashrekord

loosebazooka:
Create two bundles with the same certificate, replaced the tlog entry of one with the other. The only mismatched item is the artifact hash in the hashrekord.

Clients should be inspecting the hashrekord and validating it against the provided artifact


woodruffw (approved):

woodruffw:
LGTM, thanks @loosebazooka!

PR #170, opened 2024-12-03
add skip-cpython-release-tests

woodruffw:
This adds an input that allows the (hefty) CPython release bundle tests to be outright skipped. It also adds a selftest matrix configuration to ensure the input works as expected.

### DIFF
--- a/test/test_bundle.py
+++ b/test/test_bundle.py
@@ -11,6 +11,12 @@ from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
 from test.client import BundleMaterials, ClientFail, SigstoreClient
 from test.conftest import _MakeMaterialsByType, _VerifyBundle
 
+SKIP_CPYTHON_RELEASE_TESTS = (
+    os.getenv("GHA_SIGSTORE_CONFORMANCE_SKIP_CPYTHON_RELEASE_TESTS", "false") != "false"
+)
+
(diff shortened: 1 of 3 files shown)


bdehamer (approved):
👍

test/test_bundle.py
```diff
@@ -377,13 +383,14 @@ def test_verify_rejects_mismatched_hashedrekord(
         verify_bundle(materials, input_path)
 
 
+@pytest.mark.skipif(
+    SKIP_CPYTHON_RELEASE_TESTS, reason="CPython release bundle tests explicitly skipped"
+)
+@pytest.mark.skipif(not GITHUB_WORKSPACE, reason="GITHUB_WORKSPACE not set")
```
loosebazooka, line 389: Feels like we might not detect these tests not running if GITHUB_WORKSPACE is weirdly unset? maybe that's not a real concern?

loosebazooka (approved):

test/test_bundle.py
woodruffw: Yeah, that should only happen on local runs, where `GITHUB_WORKSPACE` should be unset anyways. And in that case we'll still present a skip warning as part of the test summary.

(I want to do a follow-on refactor that removes the `GITHUB_WORKSPACE` reference here entirely, however, since it makes local testing more confusing.)

Issue #122, opened 2024-01-24
Discussion: remove non-bundle ("detached materials") tests from the conformance suite?


woodruffw:
Opening this up for discussion, since other clients may have strong opinions here.

The context on my end: `sigstore-python` *may* begin moving towards a position of only supporting bundles on the CLI, *if* we can. The rationale for this is threefold:

1. Bundles are standardized and included in the client spec, while "detached" materials are only mentioned in passing (and more in terms of what they provide, not their format). This makes it harder to define consistent client behavior around detached materials.
1. "Detached" materials result in a much worse user experience: there are 3+ sidecar files for every input being verified instead of just one (`foo.{crt,sig,rekor}` instead of `foo.sigstore.json`), and handling them correctly results in a more complicated CLI than strictly necessary.
1. "Detached" materials complicate the handling of Rekor entries: unlike bundles where the bundle version tells us whether an entry must be supplied, clients must individually determine how to handle the case of `foo.rekor` missing, present, etc.

If this rationale is sound and amenable to other clients, then I propose the following changes to the conformance suite:

1. Existing "detached" tests should be converted or re-created as bundle tests, if they don't already have bundle equivalents;
2. The CLI "protocol" should be amended to remove `sign` and `verify` entirely, leaving only `sign-bundle` and `verify-bundle`.

On the other hand, if this is too hasty, we *could* retain the current detached tests and use the "skip" machinery to ensure that non-supporting clients can continue to use the conformance suite.

Thoughts are greatly welcome here! I'm not 100% sure this is a good idea yet; on one hand it'll simplify conformance testing substantially but on the other it's a significant change 🙂

steiza:
I'm in favor of making the conformance test suite be all bundle-based; I think bundles are the future direction we want to go in.

But I'll add another option to the mix - `sigstore-python` could only support bundles and still be able to handle the existing non-bundle conformance tests by using the detached materials to construct a bundle.

This is exactly what `sigstore-go` does, as it only understands how to verify bundles: see https://github.com/sigstore/sigstore-go/blob/2e5477d970784bb700b510402c1b1325cd2d47f6/cmd/conformance/main.go#L101-L152

woodruffw:
> But I'll add another option to the mix - `sigstore-python` could only support bundles and still be able to handle the existing non-bundle conformance tests by using the detached materials to construct a bundle.

[...]

PR #173, opened 2024-12-06
Remove conformance tests that use detached materials

steiza:
#### Summary

For #122. All the clients (including `cosign`!) are moving away from detached materials in favor of bundles.

This will help unblock things like https://github.com/sigstore/sigstore-go/pull/277.

#### Release Note

* Removed `sign` and `verify` from conformance test protocol; all tests use `sign-bundle` and `verify-bundle`.

#### Documentation

N/Aish; already updated `docs/cli_protocol.md` as part of this pull request.


woodruffw (dismissed):
Good riddance!

loosebazooka:
Could it make sense to organize these into their own test file and let people explicitly include it? Could it help prevent a regression in cosign while we are transitioning it?

woodruffw:
> Could it make sense to organize these into their own test file and let people explicitly include it? Could it help prevent a regression in cosign while we are transitioning it?

I'm OK with this, *if* we want to keep them around: IMO having the disabled by default and then allowing integrators to do `with-nonstandard-detached-material-tests: true` would probably be sufficient 🙂 

OTOH I think this is all 100% unstandardized, so having them in the conformance suite at all isn't fantastic, especially if it's dragging down efforts to get other things added. So if `cosign` et al. want these for regression testing, it might make sense to copy/vendor them directly into the test suite there.

steiza:
> Could it help prevent a regression in cosign while we are transitioning it?

Cosign's current conformance test integration uses sigstore-go for bundles, and for non-bundles it first assembles the detached materials into a bundle so it can use sigstore-go: https://github.com/sigstore/cosign/blob/e7667bd9d71d99aaa9ad3489241f5abe92fe0c9c/cmd/conformance/main.go#L114-L120

It's possible that after https://github.com/sigstore/cosign/pull/3844 lands (which is currently a draft), we could update cosign's conformance test to exercise the detached materials case. On the other hand, the conformance detached materials support is blocking other work, like https://github.com/sigstore/sigstore-go/pull/277.

loosebazooka (dismissed):

loosebazooka:
Alright let's go!

haydentherapper (dismissed):
Have we verified that all tests here are covered by the bundle tests?

steiza:
> Have we verified that all tests here are covered by the bundle tests?

There is a surprising amount of subtlety here.

I added a test back in that checks that the ctlog public key is correct. The other test on mismatched artifacts which should be covered by `test_verify_rejects_mismatched_hashedrekord`. The empty tests are checking for mismatched signatures (or certificates), which should be covered by `test_verify_rejects_invalid_signature`.

I don't think we're substantially losing coverage with this change, although the way some cases are tested are changing slightly.

haydentherapper (approved):

woodruffw (approved):

woodruffw:
Thanks a ton @steiza!

PR #176, opened 2025-01-09
handle different certificate fields correctly

woodruffw:
~~WIP; still root causing the failure that surfaced this.~~

This checks both `x509_certificate_chain` and `certificate` in the `test_sign_does_not_produce_root` test, since clients can place bundled certificates in either (and as of v3 bundles, are required to use the latter instead of the former).

xref https://github.com/sigstore/sigstore-go/actions/runs/12697653217

CC @haydentherapper

### DIFF
--- a/test/test_bundle.py
+++ b/test/test_bundle.py
@@ -105,8 +105,14 @@ def test_sign_does_not_produce_root(
     bundle = Bundle().from_json(bundle_contents)
 
     # Iterate over our cert chain and check for roots.
-    certs = bundle.verification_material.x509_certificate_chain
-    for x509cert in certs.certificates:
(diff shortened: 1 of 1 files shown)


haydentherapper (approved):

woodruffw:
Current hypothesis: other clients that were running this test were generating v2 and older bundles, which generally use `x509_certificate_chain`.

haydentherapper:
I believe this is related to bumping betterproto. As noted in https://github.com/danielgtaylor/python-betterproto/releases/tag/v.2.0.0b7, accessing an unset field not throws an error.

woodruffw:
Nice catch! Okay, this makes sense to me then!

PR #179, opened 2025-01-13
Pin dependencies, setup dependabot

jku:
Purpose here is to make the test suite reproducible for action users (test suite dependency updates should not break workflows). At the moment I believe the action is broken for everyone except sigstore-python and the selftest here...

* Pin dependencies by exact version
* Setup dependabot: update once a month

Fixes #178, but I will file another issue about installing sigstore-python in a virtual env of its own

Issue #178, opened 2025-01-13
Conformance tests failing after latest protobuf-specs python release


haydentherapper:
See https://github.com/sigstore/cosign/actions/runs/12738467371/job/35500829830 for an example. 

```
FAILED ../../_actions/sigstore/sigstore-conformance/main/test/test_bundle.py::test_sign_does_not_produce_root[ArtifactInputType.PATH] - pydantic_core._pydantic_core.ValidationError: 1 validation error for Bundle
  Value error, Group content has no value; all fields are None [type=value_error, input_value=ArgsKwargs(()), input_type=ArgsKwargs]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

cc @woodruffw

jku:
I think our requirements.txt should pin versions strictly:
* this would not add more work since we typically update requirements.txt on upgrades anyway 
* this makes using the action reproducible which I think is quite valuable

jku:
> we typically update requirements.txt on upgrades anyway

oops that's a lie. We don't have dependabot!

jku:
So debugging a little further: this issue does not show up in our self tests because
* sigstore-python is installed first (not in virtual env)
* test suite is then installed in virtual env: pip finds the protobufs already installed and is currently happy with that version...

We probably do want to start installing a sigstore-python in its own virtual env so different pinned deps are possible

jku:
Let's do a release once these are merged and look ok: I think the action is currently broken for everyone (except the selftest here and sigstore-python)

haydentherapper:
Thanks y'all for the quick fix!

haydentherapper:
Confirmed passing, thanks again! https://github.com/sigstore/cosign/actions/runs/12752366181

PR #181, opened 2025-01-13
self-test: Install sigstore in separate virtualenv

### LINKED ISSUE #180: self-test should install sigstore-python in a virtual env
self test (and possibly makefile) should probably install sigstore-python in a separate virtual environment as current setup (where sigstore-python is installed outside of the virtual environment) leads to dependency version clashes between the test suite and sigstore-python.

jku:
This prevents sigstore dependencies from being accidentally used by the test suite.

Note that Makefile  currently still installs everything in the same environment. Fixing this requires some additional work: either the documentation needs to include the use of venv when invoking the test suite manually (making the invocation even more complicated) or the sigstore-python-conformance entrypoint needs to somehow handle the virtual environment...

Fixes #180


jku:
I think the PR is now correct (if a little ugly -- the way we invoke the wrapper could be a lot more straightforward).

The failure in the selftest is correct: we now reproduce #178. We can merge #179 first, then this should start passing (since we pin protobuf-specs).

.github/workflows/conformance.yml
```diff
@@ -28,12 +28,23 @@ jobs:
           cache: "pip"
 
       - name: install sigstore-python
-        run: pip install "sigstore >= 3.5.0, < 4.0"
+        run: |
+          python3 -m venv sigstore-env
```
jku, line 32: maybe some other name would be useful sinc ethe workflow already uses sigstore-env for matrix name

.github/workflows/conformance.yml
```diff
@@ -28,12 +28,23 @@ jobs:
...
+          ./sigstore-env/bin/python -m pip install "sigstore >= 3.5.0, < 4.0"
+
+          cat <<EOF > ./sigstore-env/bin/conformance
+          #!/bin/sh
+          # allow sigstore-python-conformance wrapper to find "sigstore" binary from venv
+          PATH=$PWD/sigstore-env/bin/:$PATH
+          $PWD/sigstore-env/bin/python $PWD/sigstore-python-conformance "\$@"
+          EOF
```
woodruffw, line 40: I don't love that we're generating a script on the fly for this, but I don't have a better suggestion off the top of my head.

di:
Trying to make sure I understand the goal here: we want to install `sigstore-python`, but use the sub-dependencies specified in the `requirements.txt` here and not it's sub dependencies, correct? Wouldn't `pip install sigstore --no-deps` do this?

jku:
@di the goal is to install sigstore-python so that it does **not** poison the virtual environment for the test suite: The practical issue was that e.g. #178 does not appear in our self test but does break every user.

Current situation is:
* sigstore-python is first installed globally (no venv)
* the test suite is installed in its own venv but that finds the shared dependencies that have been installed globally

This PR changes things so sigstore-python is installed in a  separate venv. The only complication is that currently the invocation is a bit complicated (generated shell script calls the sigstore-python-conformance wrapper that calls "sigstore" binary): I'm pretty sure there is a simpler setup but it will require modifying the wrapper somehow

.github/workflows/conformance.yml
jku: I'll file an improvement issue, it's not amazing

jku:
I agree generating the script is not pretty but I'll merge as is: I think it still does the right thing and we can improve after the action is working again.

PR #187, opened 2025-01-14
Add (selftest) virtual environment management

jku:
Two goals:
* Make local testing easy when self test requirements (sigstore-python) and actual test suite requirements are different 
* Simplify self test setup in the self test workflow

Changes:
* self test requirements are separated into `selftest-requirements.txt` (similar unused `tools/requirements.txt` was removed)
* Makefile now creates a separate virtual environment `selftest-env` for selftest sigstore client
* sigstore-python-conformance expects to find the sigstore binary in `selftest-env`
* workflow is simplified similarly

The fact that sigstore-python-conformance now requires the venv `selftest-env` to exist is a bit restricting:  but it works and makes  issues like the sigstore-protobuf-specs upgrade in #194  locally debuggable -- if other options are needed later on it will be easy to add a env variable or something to override the binary location


jku:
* Rebased on main 
* Did the rename I suggested myself in pervious comment: "selftest" seems more useful as name than "sigstore"

loosebazooka (commented):
lgtm, but maybe @woodruffw should do a real review?

jku:
I think this might need changes since #192 was merged before this

loosebazooka (approved):
Does this need a readme update?

jku:
> Does this need a readme update?

I'm leaning towards No:
* the "Development" example in the README is supposed to be a somewhat generic example of running the test suite -- including implementation details of the python client-under-test seems like too much detail
* the example should just work: `make dev` creates selftest-env

PR #196, opened 2025-02-24
cpython verify: Only verify one artifact per release

jku:
This should improve performance by a lot but should not affect coverage (since all artifacts in one release are likely very similar).

Changes in the PR:
* non-sigstore signed cpython releases are not counted as subtests (this likely did not affect performance in any way but seems like a better way to count)
* only one artifact per cpython release is tested

We're left with 70 cpython release tests (down from 1633). The selftest runtime went from 330 seconds to 90 seconds, I think sigstore-java will get even more benefit from this.

### DIFF
--- a/test/test_bundle.py
+++ b/test/test_bundle.py
@@ -445,10 +445,10 @@ def test_verify_cpython_release_bundles(subtests, client):
 
         version = json.loads(version_path.read_text())
         for artifact in version:
+            bundle = artifact.get("sigstore")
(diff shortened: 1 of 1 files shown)


loosebazooka (approved):
nice

woodruffw (approved):
Makes sense to me! It might be nice to have a "full" mode where we run against all of them, but I agree it's overkill as-is 🙂

PR #216, opened 2025-06-13
Basic rekorv2 tests

jku:
Add very basic tests for verifying rekorv2 bundles (happy path test and a few timestamp related tests).
* This uses sigstore-python from main branch as the selftest client
* All tests use a custom trust root (that is really just staging trusted root since it has rekor2 in it)

Clients can set XFAIL as usual if they do not support rekorv2 yet (we don't have a SKIP input so they need to do XFAIL but maybe we should have it  so clients could just skip _all_ tests they don't support -- as an example `"*intoto*, *rekor2*"`) instead of having to XFAIL just the ones that have unexpected results


jku:
The work in #218 allows doing this in a parametrized manner: merging that first makes sense since this is still blocked by sigstore-python

jku:
The size of the PR could be cut in half by having some smarter way of using the staging trusted root as `--trusted-root` argument (the four trusted_root.json files are identical)... but I decided to keep this simple

test/assets/bundle-verify/rekor2-timestamp-payload-mismatch_fail/README
```diff
@@ -0,0 +1,16 @@
...
+
+
+Test was created by creating two bundles at same time and manually replacing the 
+timestamp from one with the other (TOKEN is token from 
+https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon):
+
+   sigstore --trust-config ~/src/sigstore-python/test/assets/trust_config/staging-but-sign-with-rekor-v2.json sign --identity-token $TOKEN --bundle bundle2.sigstore.json test/assets/bundle-verify/a.txt
+
+   sigstore --trust-config ~/src/sigstore-python/test/assets/trust_config/staging-but-sign-with-rekor-v2.json sign --identity-token $TOKEN --bundle bundle.sigstore.json test/assets/bundle-verify/a.txt
+
```
loosebazooka, line 13: I think there's a jq thing we can run to "automate" this replacement, if you want to document that (might be more complexity than worth)

```
jq --slurpfile bundle2 bundle2.sigstore.json '.verificationMaterial.timestampVerificationData = $bundle2[0].verificationMaterial.timestampVerificationData' bundle.sigstore.json > invalidBundle.sigstore.json
```

loosebazooka (approved):
@aaronlew02 do we have any other test-cases that might be interesting to add (in a followup pr)?

PR #218, opened 2025-06-16
Parametrize almost all verification tests

### LINKED ISSUE #217: refactor bundle verification tests
Adding tests with new bundles feel like quite a bit of work: I think most of the bundle verification tests could work as a single test that is parametrized over asset directories. Directory name should end with _fail if we expect failure. Each directory should contain:
* the bundle
* optional trusted_root.json (otherwise use prod)
* README: The content should be used as the explanation message for unexpected failure/success

This way adding new verification tests would be mostly just a matter of creating the bundle.

jku:
Parametrize practically all verification tests:
* There is now just `test_verify()`  that gets parametrized over all subdirs of `test/assets/bundle-verify`: more than 20 test methods are removed 
* All tests should be functionally the same as before
* There is a README on how this works
* A side-effect is that all of the affected tests change their name: XFAIL input arguments will have to be updated
* the XFAIL input now accept shell wildcards like "*intoto*"

Fixes #217 

The one caveat here is  that it's quite hard to verify that the tests work as intended -- I have checked the output of every negative test but unfortunately #153 and https://github.com/trailofbits/rfc3161-client/issues/162 mean that a lot of the tests don't really work with our selftest client


loosebazooka:
I'll try this build out with sigstore-java

README.md
```diff
@@ -99,7 +103,7 @@ The test suite can be configured with
 Following example runs the test suite with the included sigstore-python-conformance client script:
```
loosebazooka, line 103: add above running a single test with `-k` ?

test/assets/bundle-verify/README.md
```diff
@@ -0,0 +1,13 @@
...
+* Name of the directory is the parametrized part of test name
+* Name should end in "_fail" if the verification is expected to fail
+* Directory must contain
+  * `bundle.sigstore.json`: The signature bundle
+  * `README`: Explanation of the test: why it should fail/succeed
+* Directory may optionally contain
+  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent
+    directory is used)
+  * `trusted_root.json`: a custom trusted root (if one is not provided,
+    the Sigstore public good production instance is used)
```
loosebazooka, line 13: does staging modify this?

test/assets/bundle-verify/wrong-material_fail/README
```diff
@@ -0,0 +1 @@
+This is a valid bundle for the default artifact ("../a.txt"), but the artifact here is different
```
loosebazooka, line 1: am I crazy? there's no bundle in this directory?

loosebazooka (commented):
I think I might see the issues with failures not mapping to same way even if they are still failures? It's quite hard to tell.

test/assets/bundle-verify/README.md
jku: No, I think these tests should not be marked as using any env: 
* by default test uses production (client might  check prod tuf but no other services)
* with custom trust root clients should not connect to any services at all

We could even modify the default setup so that we provide the prod trust root as argument: then clients would not need a network even for the prod TUF ping.

test/assets/bundle-verify/wrong-material_fail/README
jku: oh wow, we have a very agressive `.gitignore` that hides all bundle files... Thanks, will fix

jku:
> I think I might see the issues with failures not mapping to same way even if they are still failures? It's quite hard to tell.

Yeah I think the only real way to compare is to 
* be on main, modify the negative test to expect success (to see the current failure mode)
* then be on the PR branch, rename the directory so it does not end in _fail (to see the new failure mode)
* eyeball the resulting outputs

test/assets/bundle-verify/bundle-with-root-cert_fail/README
```diff
@@ -0,0 +1 @@
+bundle certificate chain contains a root certificate: client should reject this
```
ramonpetgrave64, line 1: Why should clients reject this kind of bundle?

test/assets/bundle-verify/dsse-invalid-sig_fail/README
ramonpetgrave64, line 1: The title here is self-explanatory, but this should still have some description on the reason for failure. For example,  if the signature has the incorrect length.

test/assets/bundle-verify/intoto-expired-certificate_fail/README
ramonpetgrave64, line 1: Is this for an expired CA certificate? If so rename this case to `intoto-expired-ca-certificate_fail`.

test/conftest.py
```diff
@@ -252,5 +256,17 @@ def workspace():
...
+
+def pytest_generate_tests(metafunc):
+    """
+    Parametrize bundle_verify tests over all sudirectories under assets/bundle-verify
+    """
+    if "bundle_verify_dir" in metafunc.fixturenames:
+        asset_root = Path(__file__).parent / "assets" / "bundle-verify"
+        directories = [d for d in asset_root.iterdir() if d.is_dir()]
+        dir_paths = [str(d) for d in directories]
+        metafunc.parametrize("bundle_verify_dir", dir_paths, ids=[d.name for d in directories])
```
ramonpetgrave64, line 272: Non-blocker: it could be nicer to set the xfail here, like
```suggestion
        dir_paths = [str(d) for d in directories]
        params = [pytest.param(str(d), id=d.name, marks=[pytest.mark.xfail] if str(d).endswith("fail") else [])
        metafunc.parametrize("bundle_verify_dir", params)
```

ramonpetgrave64 (changes requested):
Looks mostly to me, but I would also pretty-print the json's to make this easier to review.

jku:
> Looks mostly to me, but I would also pretty-print the json's to make this easier to review.

I agree we should request that for new assets but I expect in this PR a lot of git renames would become large diffs...

test/assets/bundle-verify/bundle-with-root-cert_fail/README
jku: I don't want this refactor to be blocked by improvements that could have been done at anytime in the last two years, will file a new issue for these: #220

ramonpetgrave64 (approved):
Looks good, since we wiil followup in https://github.com/sigstore/sigstore-conformance/issues/220

jku:
I've taken another look after holidays and I see a possible issue in the selftest client in e.g. `test_verify[PATH-intoto-with-custom-trust-root]` (if un-xfailed): 

E               rekor=RekorClient(trust_config._inner.signing_config.tlog_urls[0]),
E                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
E           IndexError: list index out of range

that should not happen

jku:
Merging now: @loosebazooka did promise to check the results for sigstore-java (to ensure the tests that expect verification failure are failing for the right reasons) but merging now allows us to start adding rekor v2 bundles.

PR #235, opened 2025-08-07
Add timestamp-related Rekor v2 tests

aaronlew02:
Signed-off-by: Aaron Lew <64337293+aaronlew02@users.noreply.github.com>


jku (commented):
these look great, thanks.

The failure in selftests is very likely what appu identified: rfc3161-client library has been fixed but it looks like not yet released so sigstore-python does not use it yet -- I'll ping the maintainers

jku (approved):
rfc3161-client release happened lightning fast and the selftest turned green 👌 

These tests look good to me.

PR #237, opened 2025-08-12
Add test for rekor2 with inclusion proof removed

loosebazooka:
Fixes https://github.com/sigstore/rekor-tiles/issues/30


loosebazooka:
`sigstore-java` will fail with no log entry. We don't have a great validation policy definition system in it.

PR #239, opened 2025-08-14
Support --signing-config in CLI, test for rekor2 signing

jku:
Contents:
* Add `--signing-config` (and `--trusted-root`) to the client-under-test CLI when signing
* Make selftest client compatible with the new options
* Add a test that verifies clients capability to sign with rekor2 using the new flags

Background:
* We don't really want to use .e.g. `--staging` to test for specific results since we know the expected result changes over time (e.g. recommended tlog to use changes from rekor1 to rekor2)
* Instead we want to provide a specific signingconfig and trustedroot so we know the expected result -- the downside of this is that the services themselves may be turned offline at some point: this signingconfig and trustedroot needs to be updated at that point
* verifying the resulting bundle with the selftest client should mean that the bundle is good (because we know the selftest-client passes the verification tests). Running only the clients own verification tests could be problematic (imagine a client that hasn't yet implemented TSA signing or verification -- the signing test for rekor2 could then pass even if there was no timestamp in the bundle)

### DIFF
--- a/docs/cli_protocol.md
+++ b/docs/cli_protocol.md
@@ -40,6 +40,8 @@ ${ENTRYPOINT} sign-bundle [--staging] --identity-token TOKEN --bundle FILE FILE
 | `--staging`        | Presence indicates client should use Sigstore staging infrastructure |
 | `--identity-token` | The OIDC identity token to use |
 | `--bundle FILE` | The path to write the bundle to |
+| `--trusted-root TRUSTROOT` | Optional path to a custom trusted root to use to verify the bundle |
+| `--signing-config SIGNINGCONFIG` | Optional path to a custom signing config to use to verify the bundle |
 | `FILE` | The artifact to sign |
 
 ### Verify
@@ -56,5 +58,5 @@ ${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDE
 | `--bundle FILE` | The path to the Sigstore bundle to verify |
 | `--certificate-identity IDENTITY` | The expected identity in the signing certificate's SAN extension |
 | `--certificate-oidc-issuer URL` | The expected OIDC issuer for the signing certificate |
-| `--trusted-root` | The path of the custom trusted root to use to verify the bundle |
+| `--trusted-root TRUSTROOT` | Optional path to a custom trusted root to use to verify the bundle |
 | `FILE_OR_DIGEST` | The path to the artifact to verify, or its digest. The digest should start with the `sha256:` prefix, should be the right length for a hexadecimal SHA-256 digest, and should not be a path on disk. If any of those conditions are not met, the input should be interpreted as a filepath instead. |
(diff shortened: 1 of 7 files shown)


test/test_bundle.py
```diff
@@ -88,6 +89,43 @@ def test_sign_does_not_produce_root(
...
+) -> None:
+    """
+    assert that client can sign a rekor 2 bundle
+    """
+
+    materials: BundleMaterials
+    input_path, materials = make_materials_by_type("a.txt", BundleMaterials)
+    assert not materials.exists()
+
+    materials.signing_config = input_path.parent / "rekor2_signing_config.json"
```
jku, line 106: Now that we can see how we want to use the test assets, I think we should refactor the make_materials methods, likely so that each set of test assets  is in its own directory (like the bundle_verify tests already are)... so far I just did this ugly hack here as a proof of concept.

On the other hand, I think we might not need more tests for rekor2 so maybe this is not too bad: using the selftest clients verify on the bundle produced by client-under-test should guarantee that the bundle is pretty good since the selftest client passes the verify tests (this assumes that our verify tests are comprehensive -- but I think it makes sense to improve the verify tests if that is not the case)

jku:
@loosebazooka @aaronlew02 as mentioned in top comment I'm still considering a slightly deeper refactor, but if you can do a sanity check already that's appreciated:
* does the "API change" seem ok?
* do you agree with the idea that we can avoid testing every detail in the bundle manually by running the bundle through selftest-client verify -- this way the better verification tests also makes the signing test more trust worthy?

test/assets/rekor2_signing_config.json
haydentherapper, line 1: Could these two files be dynamically fetched from a TUF repo before passing them to the client, rather than hardcoding them, to avoid them getting out of date? Do we have anything like this for the verification path?

You mentioned not needing to use `--staging`, but wouldn't we want to test against both production and staging going forward? That flag could determine which TUF repo we pull from.

docs/cli_protocol.md
```diff
@@ -40,6 +40,8 @@ ${ENTRYPOINT} sign-bundle [--staging] --identity-token TOKEN --bundle FILE FILE
 | `--staging`        | Presence indicates client should use Sigstore staging infrastructure |
 | `--identity-token` | The OIDC identity token to use |
 | `--bundle FILE` | The path to write the bundle to |
+| `--trusted-root TRUSTROOT` | Optional path to a custom trusted root to use to verify the bundle |
+| `--signing-config SIGNINGCONFIG` | Optional path to a custom signing config to use to verify the bundle |
```
loosebazooka, line 44:
```diff
- | `--signing-config SIGNINGCONFIG` | Optional .... to use to verify the bundle |
+ | `--signing-config SIGNINGCONFIG` | Optional .... to define signing endpoints |
```

or something like that

loosebazooka (dismissed):

jku:
The documentation tweak also made the tests rerun: this proves the test works after yesterdays log2025-alpha1.rekor.sigstage.dev shard freeze :)

haydentherapper (approved):
Awesome to see this indirectly test how clients handle frozen shards too!

PR #246, opened 2025-09-02
Change order of arguments for sign bundle test

Hayden-IO:
--trusted-root and --signing-config were appended after the artifact argument. A conformance test client implementing the CLI protocol would sign the last argument, meaning it would sign the signing config itself, which would cause an error during verification.

This appends the artifact to the argument list last.


jku:
FYI, I also realised there is an issue with my use of the selftest client in the signing test (I still think it's a decent idea but the action install needs a tweak)

Issue #243, opened 2025-08-22
filling in the testing gaps


loosebazooka:
Now that aaron converted java conformance as a server, we can run easy code coverage on conformance over java looking for gaps in our testing, things we *could* add

### bundle
- [ ] fail if a bundle with no dsse or hashedrekord
- [ ] fail on empty inner timestamp data (`timestamp.rfc3161timestamp`)
- [ ] unexpected types? `mashedrekor:1.0.0` `dsse:3.0.0`
- [ ] certificate id/issuer matching? or other fields

### hashedrekord
- [ ] mismatched signature (actual vs logged)
- [ ] mismatched hash (actual vs logged)
- [ ] mismatched cert (actual vs logged)

### dsse
- [ ] unsupported payload type
- [ ] provided artifact hash is not in subjects
- [ ] fail if multiple signatures provided (sigstore limitation)
- [ ] mismatched cert (not sure if this is actually reachable)

### checkpoints
- [ ] fail if more than 20 sigs
- [ ] fail if no matching signature found

There's more tests related to unsupported algorithms that would still be useful to check that we don't fail open when we hit unparseable information.

aaronlew02:
The checkpoint test for no matching signature has been added in #250. The test for excessive signatures will not be implemented in conformance due to the subjectivity of such a check (see comments in #250).

loosebazooka:
@loosebazooka @aaronlew02 we might need to check if this is complete actually

PR #250, opened 2025-09-04
Add checkpoint signature mismatch test for Rekor v2

aaronlew02:
Partially addresses #243 

#### Summary

This change adds a test for a Rekor v2 bundle in which there is no checkpoint signature matching the identity of the transparency log.


haydentherapper:
My two cents, I'd prefer to not test for an excessive number of signatures. Clients _should_ have some limits in place, but this is true for any repeated field to prevent a DoS. Hardcoding 20 as a limit seems too prescriptive.

Testing no signature SGTM.

haydentherapper (approved):

jku (approved):
thanks, looks good to me

PR #254, opened 2025-09-24
Self-test on all supported platforms

haydentherapper:
This will give us an early signal if there's a platform-specific issue.


haydentherapper:
Well I thought this would take a second, but of course with Windows it's not. Any idea what needs to be done?

jku:
Can you write an issue or a description here -- what is the goal here? Is it that conformance should support all major platforms and that we should advice all users to run the full test suite on all the platforms their client works on?

PR #264, opened 2025-10-29
Use platformdirs for cache directory path

aaronlew02:
#### Summary

This change replaces the hardcoded cache directory path used for the `sigstore-python` cache with a path dynamically generated by `platformdirs` to accommodate various operating systems.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -3,3 +3,4 @@ pytest-subtests==0.15.0
 requests==2.32.5
 cryptography==46.0.3
 sigstore-protobuf-specs==0.5.0
+platformdirs==4.4.0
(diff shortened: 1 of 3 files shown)


loosebazooka:
So, what's the correct way to add the platformdirs dep, an earlier run on ci failed: https://github.com/sigstore/sigstore-conformance/actions/runs/18917719327

jku:
> So, what's the correct way to add the platformdirs dep, an earlier run on ci failed: https://github.com/sigstore/sigstore-conformance/actions/runs/18917719327

This is the correct way... it's just that the result of generating `requirements.txt` can depend on python version.
* I could generate requirements.txt on a (more) supported Python version (or trigger dependabot after this merges) but that likely means it stops working for Aaron again...
* We could theoretically maintain multiple requirements lists and test the action on multiple python versions. but that's a lot of work just to allow developers to run older python
* From the perspective of the action we should just enforce a single python version -- but that's a little harsh for the developer experience unless we provide a python-installation mechanism in Makefile (we're pretty close to being able to do that with uv but it's not trivial)

di:
> From the perspective of the action we should just enforce a single python version -- but that's a little harsh for the developer experience unless we provide a python-installation mechanism in Makefile (we're pretty close to being able to do that with uv but it's not trivial)

I think this would be reasonable to do -- this is closer to an application than a library. We could set `.python-version` here and guide folks to using https://github.com/pyenv/pyenv, or using `uv` should also be sufficient.

jku (approved):

PR #283, opened 2025-11-11
feat: Include conformance action version in client report

di:
This adds the version of the conformance suite used to the results table.

### DIFF
--- a/.github/scripts/generate_client_report.py
+++ b/.github/scripts/generate_client_report.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


action.py
```diff
@@ -55,6 +56,16 @@ def _sigstore_conformance(environment: str) -> int:
...
+# Inject the action version into the report
+with open("conformance-report.json", "r+") as f:
+    report_data = json.load(f)
+    report_data["conformance_action_version"] = os.getenv(
+        "GHA_SIGSTORE_CONFORMANCE_ACTION_VERSION", "unknown"
+    )
+    f.seek(0)
+    json.dump(report_data, f)
+    f.truncate()
+
```
jku: this works... I believe there is a neater way to inject this though:

You can provide an argument  to pytest, something like:
```python
    version = os.getenv("GHA_SIGSTORE_CONFORMANCE_ACTION_VERSION", "unknown")
    args.append(["--metadata", "conformance_action_version", version])
``` 

This should add a `environment.conformance_action_version` field into the report so the parsing needs a change as well.

jku (dismissed):

PR #284, opened 2025-11-11
feat: Add links to client SHA and workflow run

di:
This adds the version of the client (as well as links to the SHA and workflow run) to the results table.


jku (commented):
Am I missing something or are these variables not available to us without making the user provide them?

I mean that I imagined we could just look at the underlying variables `GITHUB_SHA` etc here

jku (approved):
Thanks, looks great

Issue #287, opened 2025-11-14
conformance report generation issues


jku:
#283 and #284 added some data to the report but there is an issue and the report does not get properly filled. I'm looking at the results in current selftest runs and trying the generate_client_report.py on that manually

* variables like GHA_SIGSTORE_CONFORMANCE_CLIENT_SHA can't be used in the generate_client_report.py (since that runs in our report workflow, not in the clients conformance workflow). The variables need to be used in action.py, similarly to GHA_SIGSTORE_CONFORMANCE_ACTION_VERSION is used, and then need to be parsed from the results by generate_client_report.py 
* In addition, GHA_SIGSTORE_CONFORMANCE_ACTION_VERSION is not getting inserted into the result json like I expected it to (with "pytest --metadata").

The second issue I don't understand yet, will look into it a bit

CC @di

jku:
https://github.com/numirias/pytest-json-report/issues/89 -- it looks like pytest-json-report has not had a release (or other work) in a while so has been incompatible with pytest-metadata for some time now. `pytest-metadata==2.0.4` may work but then we are using two outdated packages...

EDIT: I tested and `pytest-metadata==2.0.4` does work

PR #288, opened 2025-11-14
Pass client_sha and client_sha_url as pytest metadata

di:
Part of #287 (although if `--metadata` is not working, this will not work until we figure out what's going on there).

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,6 +1,5 @@
 pytest==8.4.2
 pytest-json-report==1.5.0
-pytest-metadata==3.1.1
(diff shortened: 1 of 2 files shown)


jku (commented):
An annoying situation: pytest-json-report is not really maintained -- and as a result it's not compatible with current pytest-metadata

some potential solutions:
* just fall back to your original idea of injecting data into the file after it's written by pytest
* pin `pytest-metadata==2.0.4` -- this does mean we have two very old dependencies (both are very simple ones though)

I think I'm ok with either one

PR #306, opened 2025-12-17
suggest implementation strategy for managed key verify tests

### LINKED ISSUE #301: Add support for key-based verification
**Description**

At the recent clients and TSC meeting, we determined we would like conformant Sigstore client libraries to be able to verify protobuf bundles that were signed with managed keys, in addition to the verification we support today of signing certificates from Fulcio obtained with an OIDC token.

There's a couple of steps needed for this:

1. We need to update the [conformance CLI verify protocol](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md) to include verifying with a provided public key, instead of a certificate identity and certificate OIDC issuer. We could take inspiration from / learn lessons from https://github.com/sigstore/sigstore-go/tree/main/examples/sigstore-go-verification here. Note that we should continue to accept a trusted root when verifying with a public key, in case the bundle contains transparency log entries or signed timestamps that also need to be verified.

2. Once we agree on and land the update to the conformance CLI verify protocol, we can update https://github.com/sigstore/sigstore-go/blob/main/test/conformance/main.go to support the protocol (we might want to do this in parallel to test out the proposed protocol changes).

3. We should add some conformance tests that use the CLI verify protocol with managed keys, and either protect them with a flag (like `--skip-signing`) that's off by default, or include `xfail` instructions in the subsequent conformance test release.

4. Other clients can then add managed key verification support, and test their implementation with the new conformance tests.

Note that out of scope for these changes are verifying with a public key that's embedded in a X.509 certificate (although we may choose to add that at a later date). The way cosign / sigstore-go handle verifying with a user-managed certificate is to check it against the `certificateAuthorities` certificate chain in the trusted root that usually points to a Fulcio instance.

Also out of scope is having all client libraries sign with managed keys. There's a lot of complexity here around supporting various cloud HSMs (usually via REST APIs), as well as local HSMs (usually via PCKS #11). Cosign supports these use-cases today, and we encourage Sigstore users who want this functionality to use Cosign.

piceri:
#### Summary
Resolves #301 

This pull request adds testing of user managed keys to the conformance test suite. Instead of passing `--certificate-identity` and `--certificate-oidc-issuer`, these tests will instead pass `--key` with a path to the public key file. The PR contains the following test cases:

- Happy path
- No key
- Wrong key
- Key and trusted root

#### Release Note

Added tests for verification using user managed keys

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -41,4 +41,4 @@ jobs:
           entrypoint: ${{ github.workspace }}/sigstore-python-conformance
           skip-cpython-release-tests: ${{ matrix.skip-cpython-release-tests }}
           environment: ${{ matrix.sigstore-infra }}
-          xfail: "test_verify*intoto-with-custom-trust-root]"
+          xfail: "test_verify*-intoto-with-custom-trust-root] test_verify*managed-key-happy-path] test_verify*managed-key-and-trusted-root]"
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ The test suite can be configured with
 Following example runs the test suite with the included sigstore-python-conformance client script:
 ```sh
 (env) $ # run all tests
-(env) $ GHA_SIGSTORE_CONFORMANCE_XFAIL="test_verify*-intoto-with-custom-trust-root]" \
+(env) $ GHA_SIGSTORE_CONFORMANCE_XFAIL="test_verify*-intoto-with-custom-trust-root] test_verify*managed-key-happy-path] test_verify*managed-key-and-trusted-root]" \
     pytest -v --entrypoint=sigstore-python-conformance
 ...
 (env) $ # run single test


piceri:
> I'm sorry for being so slow on this.
> 
> I think we still need one more test, where the bundle contains a key but no key was provided to the verifier. And it would be expected to fail.
> 
> Otherwise this looks good to me? Perhaps a python person (@jku @di @woodruffw ?) should do a last check over it though for pythoniness

Good suggestion. Thanks for your help and patience on my first issue. I have added this test.

loosebazooka:
oh right, since the python client doesn't handle non-identity cert based flows, I think this needs to actually xfail some of these new tests with the self test: https://github.com/sigstore/sigstore-conformance/blob/c301daf57a16a242e7b9654f3f1ecc9a853c68a9/.github/workflows/conformance.yml#L41

jku (changes requested):
setting this to request changes just to remove my slightly too early approval: I'd like to hear that this has been tested manually with a client that passes before merging

PR #312, opened 2026-01-11
add a test for message digest mismatch with artifact hash

wolfv:
Found this while fuzzing in sigstore-rust - probably a good idea to have a conformance test.

### DIFF
--- a/test/assets/bundle-verify/message-digest-mismatch_fail/README
+++ b/test/assets/bundle-verify/message-digest-mismatch_fail/README
@@ -1,2 +1,5 @@
 The messageDigest in the bundle does not match the artifact hash.
-Verification must fail because the digest is inconsistent with the artifact.
+Verification should fail because the digest is inconsistent with the artifact.
+
+The digest field in the bundle is documented as an unauthenticated hint: this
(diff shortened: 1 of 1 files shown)


loosebazooka:
So is this test?
1. create a bundle for some artifact `c.txt` that is not `a.txt`: `c.txt.sigstore.json`
2. verify `a.txt` against `c.txt.sigstore.json`
3. expect failure

wolfv:
The test is munging the bundle.

Specifically, the `message-digest-mismatch_fail` test works like this:

1. Artifact file (a.txt) has SHA256 hash: a0cfc71271d6e278...
2. The bundle's messageDigest field has been modified to: a1cfc71271d6e278... (first byte changed from 0xa0 to 0xa1)
3. The Rekor tlog entry still has the correct hash that matches the actual file

So the test is:

- verify a.txt against a-munged-bundle.sigstore.json
- expect failure because messageDigest in bundle doesn't match artifact

This tests that verification code properly validates the bundle's messageDigest field against the computed hash of the actual artifact being verified, rather than just trusting the digest in the bundle.

PR #315, opened 2026-01-12
feat: add a simple corrupted inclusion proof check

wolfv:
Add a simple check where a bit is flipped in one of the inclusion proof hashes (invalid merkle tree).


wolfv:
Done!

PR #318, opened 2026-01-13
feat: add cve-2022-36056 check

wolfv:
This adds what I hope is a reproducer for the CVE. It did pass with a local build fo cosign 3.0.3!


woodruffw:
Someone should sanity check me, but I *think* we might already have coverage for this via `wrong-hashedrekord-entry_fail` 😅

loosebazooka:
> This looks good for a totally different entry, though to be thoroug I think we can add three more tests that change only the entry digest (sign something else with the same key), entry cert (sign the same artifact with another key, though this will also force the signature to be different), or entry sig (sign the same artifact with the same key again).

I'll have to look again, but I thought I wrote some of these tests already

jku:
I believe Appu is right about those other tests existing already (but I appreciate if you double check). I'll close this one based on the discussion

PR #319, opened 2026-01-15
Add bundle with SCT with extensions to tests

cmurphy:
TesseraCT is the tiles-backed CT log that the Sigstore infrastructure is moving to. It produces SCTs in the same format as the previous CT log did, but the extension field is now populated with the leaf index of the entry. Ensure clients don't fail to verify such an SCT.

Relates to https://github.com/sigstore/rekor-tiles/issues/73


Hayden-IO:
LGTM. We might have to hold off on merging until the Python client fix is in since it does a self test.

woodruffw (approved):
LGTM, probably needs a hash bump on sigstore-python 🙂

jku (approved):
Thank you colleen, looks great

PR #321, opened 2026-01-16
README: Bump example versions

jku:
Update action versions in our examples


jku:
> Or maybe a section of required tests and optional features (intoto, manged-key signing)

Yeah something like that would be good: it would be best if we had a real and usable "skip" for features but just documenting how to use xfail to accomplish that would make sense

PR #325, opened 2026-01-28
Add support for signing in-toto statements for DSSE envelopes

### LINKED ISSUE #145: DSSE Signing Support
**Description**

When constructing a Sigstore bundle, you have to choose if your content is going to be a Message Signature or a DSSE Envelope, which can then influence other things like if your Rekor entry type is hashedrekord or DSSE.

Today, the conformance signing tests assume you're always using a Message Signature (we do have a [verification test for bundles with a DSSE envelope](https://github.com/sigstore/sigstore-conformance/blob/7a44d6560fe6085392e98f098403a2ac5e415344/test/test_bundle.py#L33-L44).

To support this, we might need to modify the [sign-bundle CLI protocol](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md#bundle-flow) to include a boolean flag when you want a DSSE Envelope (sigstore-go's signing example uses `--in-toto` as often these are in-toto documents).

Depending on client support, we might also want to make these new DSSE signing tests opt-in to start.

aaronlew02:
Closes #145

#### Summary

This change enables testing a client's ability to sign an in-toto statement to produce a bundle containing a DSSE envelope.

#### Release Note

* Added `--in-toto` flag to the `sign-bundle` CLI protocol to support testing of DSSE signing.
* Added `test_sign_verify_dsse` to verify that clients produce valid DSSE envelopes when requested.
* Added support for `test_sign_verify_dsse` in the `selftest-client`.


jku:
> Would be nice to get the sigstore-python-conformance (selftest) support

let me know if you'd rather not do this -- I can give it a try

sigstore-python-conformance
```diff
@@ -114,8 +127,101 @@ with NamedTemporaryFile(mode="wt") as temp_file:
...
+
+            # Direct library call
+            with open(input_file, "rb") as f:
+                statement_bytes = f.read()
+
+            statement = Statement(statement_bytes)
+
+            trust_config_obj = ClientTrustConfig.production()
+            if "--staging" in sys.argv:
+                trust_config_obj = ClientTrustConfig.staging()
```
jku: there's also the case where we have a trustconfig from args... you should be able to handle that relatively cleanly by using the trust config filename around line 124

sigstore-python-conformance
```diff
@@ -11,6 +11,20 @@ import os
...
+
+# Ensure we are running inside the selftest virtualenv
+if os.path.abspath(sys.executable) != os.path.abspath(selftest_env_python):
+    if not os.path.exists(selftest_env_python):
+        # Fallback: Proceed with current execution if selftest env is missing
+        pass
+    else:
+        # Re-execute using the virtualenv python
+        target_script = os.path.abspath(__file__)
+        os.execv(selftest_env_python, [selftest_env_python, target_script] + sys.argv[1:])
```
jku: I have two suggestions for this: 
1. Use uv scripts to avoid manually handling selftest venv altogether 
2. Use `#!/usr/bin/env -S /bin/sh -c '"$(dirname "$0")/seltest-env/bin/python3" "$0"` as shebang: it's not pretty and still requires  selftest-env/ but it should work

Option 1 would be great (I've been planning to do that anyway)... unfortunately I've not yet figured out how to make uv scripts work with google pypi proxy situation so how about option 2?

test/test_bundle.py
```diff
@@ -196,3 +200,30 @@ def temp_bundle_path(bundle: dict) -> Path:
...
+    # Parse the output bundle.
+    bundle_contents = materials.bundle.read_bytes()
+    bundle = Bundle.from_dict(json.loads(bundle_contents))
+
+    # Ensure DSSE envelope is present and message signature is NOT present
+    assert bundle.is_set("dsse_envelope")
+    assert not bundle.is_set("message_signature")
+
+    # Verify the bundle
+    client.verify(materials, input_path, dsse=True)
```
jku: How about this addition:

```suggestion
    # Verify the bundle
    client.verify(materials, input_path, dsse=True)

    # Use selftest client verify to assert that the bundle is correctly formed
    selftest_client = SigstoreClient(
        str(project_root / "sigstore-python-conformance"), client.identity_token, client.staging
    )
    selftest_client.verify(materials, input_path)
```

The idea here is that client-under-test could be doing both sign and verify incorrectly (and getting a false positive in the test) but this ensures the selftest client agrees that the produced bundle is valid

jku (commented):
Looks  nice and works here too. I may have left too many comments: push back if you think some are future work.

On using direct API for the verify case: you could use `sigstore attest` as well but I guess either one is fine

docs/cli_protocol.md
```diff
@@ -49,14 +50,15 @@ ${ENTRYPOINT} sign-bundle [--staging] --identity-token TOKEN --bundle FILE FILE
 #### Bundle flow
 
 ```console
-${ENTRYPOINT} verify-bundle [--staging] --bundle FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] FILE_OR_DIGEST
+${ENTRYPOINT} verify-bundle [--staging] [--dsse] --bundle FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] FILE_OR_DIGEST
```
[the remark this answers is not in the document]
jku: this seems fair (at least assuming we find an agreement on the other question WRT dsse flow)

* for signing "--dsse" still make sense, I think
* for verify, the python wrapper could do just a bit of light parsing/grepping to decide which path to use

sigstore-python-conformance
```diff
@@ -114,8 +127,101 @@ with NamedTemporaryFile(mode="wt") as temp_file:
...
+                policy=verification_policy,
+            )
+
+            # Also check that the input file matches the statement in the bundle
+            with open(input_file, "rb") as f:
+                expected_statement = f.read()
+
+            if payload != expected_statement:
+                print("Error: Bundle payload does not match input statement", file=sys.stderr)
+                sys.exit(1)
```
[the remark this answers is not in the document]
aaronlew02: This has been moved to the test.

jku:
I'll merge this now: This makes it easier to test this with client projects, we can still make changes here in conformance before a release if needed.

PR #357, opened 2026-04-16
New testing token + refactoring

### LINKED ISSUE #240: refactor make_materials*() helpers
The way we build test cases has changed a bit through the lifetime of this project so the current helpers are not as useful as they could be -- there's likely also unused code in there.

* Practically all verification tests are now implemented in their own directories in test/assets/bundle-verify (directory contains a bundle, optional trust root, and the expected result is encoded in directory name)
* The remaining tests use varying mechanisms of material collection

I think we could refactor the `make_materials*()` helper methods so that they handle this directory-per-test-case and then build all remaining test cases in that way too.

### LINKED ISSUE #358: Testing token is constantly expired
_(moved from https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/issues/59)_

## Problem

Something seems to have changed in GitHub Actions: [trigger-extremely-dangerous-oidc-beacon.yml](https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/blob/main/.github/workflows/trigger-extremely-dangerous-oidc-beacon.yml) is maybe getting  rate limited more now and the published token is often expired, and that leads to conformance runs failing way too often.

The core issue is that the GitHub actions OIDC token is so short lived that we don't seem to be able to run a GitHub actions workflow often enough to reliably make a live token available.

I think we've spent enough time trying to make this work on Github, it's clearly not the correct platform for this.

### Required Cloud APIs

- **Cloud Run** to host the job that gets a token and publishes to GCS
- **Cloud Scheduler** to trigger the run every 15 minutes
- **Cloud Storage** to serve the token to the public

### IAM configuration

Need two service accounts so that we are able to leak a token from an identity that has no permissions
  * _leaked identity_: this identity has no permissions and should be named so that it's obvious that it should never have any
  * _runner identity_: 
    * `roles/storage.objectUser` for the GCS bucket
    * `roles/iam.serviceAccountTokenCreator` for the _untrusted identity_
    * `roles/run.invoker` for the Cloud Run Job

### GCS bucket

Needs to be public:
* `Public Access Prevention: Off`
* permissions `allUsers: roles/storage.objectViewer`
* No CloudCDN
  
### Cloud Run job
We can handle this without even building an image (as long as we're fine with a slightly obnoxious container command).

  * Container `gcr.io/google.com/cloudsdktool/google-cloud-cli:stable`
  * service account: `<email of the runner identity>`
  * Environment:
    * `UNTRUSTED_SA: <email of the leaked identity>`
    * `BUCKET: <GCS bucket name>`
  * Command
    * `bash`
    * `-c`
    * `bash set -eo pipefail; gcloud auth print-identity-token --impersonate-service-account=$UNTRUSTED_SA --audiences=sigstore --include-email | tr -d '\n' | gcloud storage cp - gs://$BUCKET/untrusted-testing-token.txt --cache-control="no-store, max-age=0" --content-type="text/plain"`
  * Job trigger: 
    * schedule  `*/15 * * * *`
    * service account `<email of the runner identity>`

jku:
I wanted to try switching the  testing token to something more reliable, but ended up doing several refactorings because they just made things simpler...

* Stop using the extremely-dangerous-public-oidc-beacon, use a Google Cloud run token leaker instead
* replace complicated expiry checks with much simpler ones -- the new token should be valid for minimum 45 minutes, enough for any test run, so there's no need to re-download during a run
* Refactor `VerificationMaterials` to support identity, issuer, key and artifact. Make sure the test parameterization is updated (so different identities and issuers can be used in `test/assets/bundle-verify`)
* refactor `test_simple()`: Verify with self test client as well (same as rekor 2 signing test already does: this ensures broken verification does not lead to false positives in signing)
* Avoid hard coding the signing identity when signing: instead `SigstoreClient.sign()` figures out the identity from the token, and updates the verification materials. This way if the test runs `SigstoreClient.verify()` on the same materials, the identity is correct as if by magic

This should require no changes from action users.


loosebazooka:
> Remove now unneeded expiry checks -- the new token should be valid for minimum 45 minutes, enough for any test run

Do we have some check here? I would be confused if my test failed and I didn't get "Expired token" logged somewhere?

jku:
> Do we have some check here? I would be confused if my test failed and I didn't get "Expired token" logged somewhere?

I would expect the client to print something like that?

But I suppose you are right: if the token leaker broke and the token is expired to begin with, we shouldn't even try to run signing tests. I think I can add the check in the client where it parses the token anyway

EDIT: Added check for expiry and some checks for malformed tokens

jku:
I'm marking this ready for review: the token now comes from a GCP project that is maintained with the rest of the public good instance projects so should be pretty reliable -- obviously I have no data to prove that it is more reliable: I think we should just try it out

test/assets/bundle-verify/README.md
```diff
@@ -7,8 +7,9 @@ Each subdirectory is used to parametrize test_verify():
   * `bundle.sigstore.json`: The signature bundle
   * `README`: Explanation of the test: why it should fail/succeed
 * Directory may optionally contain
-  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent
-    directory is used)
+  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent directory is used)
+  * `identity`: file contents are the signing identity (if not provided, "https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/.github/workflows/extremely-dangerous-oidc-beacon.yml@refs/heads/main" is used)
+  * `issuer`: file contents are the signing identity (if not provided, "https://token.actions.githubusercontent.com" is used
```
loosebazooka:
```suggestion
  * `issuer`: file contents are the signing issuer (if not provided, "https://token.actions.githubusercontent.com" is used
```

test/client.py
```diff
@@ -147,26 +199,35 @@ def raises(self):
...
 
     @sign.register
-    def _sign_for_bundle(self, materials: BundleMaterials, artifact: os.PathLike, dsse: bool = False) -> None:
+    def _sign_for_bundle(self, materials: BundleMaterials, dsse: bool = False) -> None:
         """
         Sign an artifact with the Sigstore client, producing a bundle.
 
         This is an overload of `sign` for the bundle flow and should not be called directly.
         """
+        lifetime = int((self.expiry - datetime.now()).total_seconds())
```
loosebazooka, line 223: are these guaranteed to be the same timezone? I don't know if I understand fully how datetime works in python? Does line 158 convert the payload_json timezone to local timezone?

test/conftest.py
facutuesca: this can be removed now, right?

loosebazooka (commented):
this is maybe too python for me. I'll take a look again but it might be slow.

loosebazooka:
Gemini suggested getting rid of the [`requests` import](https://github.com/sigstore/sigstore-conformance/blob/main/requirements.in#L4) in requirements.txt because it apparently isn't used ([removed here?](https://github.com/sigstore/sigstore-conformance/pull/134/changes#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2L13)). urllib3 is a transitively imported dependency but we appear to use it explicitly here -- so import it directly or use the older urllib api's to make GET calls on the token?

test/client.py
jku: yes, tz=None by default in both cases so local time

jku:
rebased on main, added the three last commits to address review comments

test/assets/bundle-verify/README.md
```diff
@@ -7,8 +7,9 @@ Each subdirectory is used to parametrize test_verify():
   * `bundle.sigstore.json`: The signature bundle
   * `README`: Explanation of the test: why it should fail/succeed
 * Directory may optionally contain
-  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent
-    directory is used)
+  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent directory is used)
+  * `identity`: file contents are the signing identity (if not provided, "https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/.github/workflows/extremely-dangerous-oidc-beacon.yml@refs/heads/main" is used)
+  * `issuer`: file contents are the signing identity issuer (if not provided, "https://token.actions.githubusercontent.com" is used
```
loosebazooka:
```suggestion
  * `issuer`: file contents are the signing identity issuer (if not provided, "https://token.actions.githubusercontent.com" is used)
```

loosebazooka (dismissed):

loosebazooka (approved):

Issue #105, opened 2023-10-11
Move bundles to v0.2


loosebazooka:
Just thinking about this problem since protobuf-specs says about v0.1 bundles: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L62

There aren't really too many compatibility issues but maybe v0.1 support should be optional for clients in conformance tests?

haydentherapper:
Agreed. We should have some tests for v1 for clients that care, but I don’t think we should mandate v1 support.

woodruffw:
Fully agreed as well!

woodruffw:
#109 starts on this as a bugfix, but others need to be moved as well.

segiddins:
I think we also need a v0.3 test as well

loosebazooka:
@loosebazooka see if we have all bundles types, perhaps add explicit checks for all versions.

loosebazooka:
It does look like we need a happy path for each bundle version.
`happy-path-v0.1` ...

we do have all bundle types in here, but they're spread out randomly, and some not even in passing tests.

<details open>

<summary> expand/collapse </summary>

| bundle version | test name |
| -------- | -------- |
| vnd.dev.sigstore.bundle+json;version=0.1  | bundle-empty-certificate-chain_fail |
|                                           | bundle-from-wrong-instance_fail |
|                                           | bundle-invalid-base64-signature_fail |
|                                           | bundle-malformed-json_fail |
|                                           | bundle-negative-log-index_fail |
|                                           | bundle-with-root-cert_fail |
|                                           | checkpoint-bad-keyhint_fail |
|                                           | checkpoint-wrong-roothash_fail |
|                                           | happy-path |
|                                           | incorrect-public-key_fail |
|                                           | integrated-time-in-future_fail |
|                                           | invalid-checkpoint-signature_fail |
|                                           | invalid-ct-key_fail |
|                                           | signature-mismatch_fail |
|                                           | wrong-material_fail |
| vnd.dev.sigstore.bundle+json;version=0.2  | intoto-expired-certificate_fail |
|                                           | intoto-log-entry-mismatch_fail |
|                                           | intoto-missing-inclusion-proof_fail |
|                                           | intoto-set-outside-signing-cert-validity_fail |
|                                           | intoto-tsa-timestamp-outside-cert-validity_fail |
|                                           | intoto-with-custom-trust-root |
|                                           | invalid-inclusion-proof_fail |
[...]

PR #359, opened 2026-04-16
Add simple validation for all bundle versions

loosebazooka:
part of #105 

- This adds happy path validation for all versions of the bundle mediaType.
- It does not do any other format checks (like a v2 bundle with invalid field from v3 -- though I'm not sure they are necessary)?


test/assets/bundle-verify/happy-path-v0.1/.bundle.sigstore.json.swp
jku, line 1: this seems unnecessary

jku (changes requested):
lgtm otherwise but drop the .swp file

test/assets/bundle-verify/happy-path-v0.1/.bundle.sigstore.json.swp
loosebazooka: Oh man new machine headaches. I'll fix it

jku (approved):

PR #364, opened 2026-04-27
Add statement attribute to BundleMaterials for DSSE signing

aaronlew02:
#### Summary

This change adds a `statement` path attribute to `BundleMaterials` to represent the statement used for signing of DSSE attestations.

Previously, `BundleMaterials` contained only a single `artifact` attribute, which, for DSSE attestations, was populated with a statement for signing and then swapped with a subject for verification. Now, the DSSE signing and verification test populates the new `statement` field with the statement and the `artifact` field with the subject, and `sign` uses the former in the DSSE case.

### DIFF
--- a/test/assets/bundle-verify/README.md
+++ b/test/assets/bundle-verify/README.md
@@ -7,7 +7,7 @@ Each subdirectory is used to parametrize test_verify():
   * `bundle.sigstore.json`: The signature bundle
   * `README`: Explanation of the test: why it should fail/succeed
 * Directory may optionally contain
-  * `artifact`: The signed artifact (if not provided, "a.txt" from the parent directory is used)
+  * `artifact`: The signed artifact. For DSSE (in-toto) envelopes, this file represents the subject of the attestation. (If not provided, "a.txt" from the parent directory is used)
   * `identity`: file contents are the signing identity (if not provided, "https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/.github/workflows/extremely-dangerous-oidc-beacon.yml@refs/heads/main" is used)
   * `issuer`: file contents are the signing identity issuer (if not provided, "https://token.actions.githubusercontent.com" is used)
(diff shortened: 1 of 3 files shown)


jku:
I was thinking of exactly this while refactoring... although I imagined we could get rid of all of the `dsse` arguments, and not add them. I figured that VerificationMaterials now includes all of the info that signing or verification needs --  am I missing something?

aaronlew02:
> I imagined we could get rid of all of the dsse arguments, and not add them

That's correct. I've removed all `dsse` boolean arguments in favor of `sign` and `verify` using the `statement` or `subject`, respectively, if available, as the artifact.

jku (changes requested):
Am I correct that we could actually avoid handling "subject" explicitly and just keep saying that in the dsse case "artifact" is the subject? I noticed this because I was surprised the `rekor2-dsse-happy-path` kept succeeding in this PR even though the `BundleMaterials.from_dir()` does not read "subject" -- I thought verify would not work but it just uses artifact instead (which in that case is the default a.txt which is the subject in this case).

I think we should either
* expect `BundleMaterials.artifact` to be subject in the dsse verify case and document that the artifact file should be a DSSE subject file inthe DSSE case in test/assets/bundle-verify/README.md (and remove `BundleMaterials.subject` to not confuse people) or
*  actually read BundleMaterials.subject from a file in from_dir(), and add a subject file  in `rekor2-dsse-happy-path` test case so it's more explicit

I'm leaning towards the former but either one works (also open to other ideas)

aaronlew02:
> Am I correct that we could actually avoid handling "subject" explicitly and just keep saying that in the dsse case "artifact" is the subject?

Yes! I had distinguished them for semantic precision, but avoiding a new `subject` field is much cleaner.

> I'm leaning towards the former but either one works (also open to other ideas)

I've gone ahead with the former approach. The latter approach is a good consideration for the future, but since there is currently only one DSSE test, the former seems more appropriate.

jku (approved):
Nice.

PR #371, opened 2026-05-13
test bundles: DSSE-with-hashedrekord on Rekor v2

codysoyland:
## Summary

Regenerates the four `rekor2-dsse-*` fixtures so the suite has end-to-end test coverage for the DSSE-with-hashedrekord client behaviour described in https://github.com/sigstore/architecture-docs/pull/63.

For DSSE envelopes the tlog entry is now a `hashedrekord` whose `data` digest is `Hash(PAE(payloadType, payload))` — the same bytes the DSSE signature covers — rather than a `dsse`/`intoto` entry over the whole envelope. The architecture-docs spec change scopes this proposal to **Rekor v2 only**, so these bundles are signed against staging Sigstore (staging Fulcio + staging Rekor v2) and their tlog entries are `kind=hashedrekord/0.0.2`, replacing the prior `dsse/0.0.2` entries. Each dir still ships its staging `trusted_root.json`; once production has a Rekor v2 instance the custom trust root can be removed.

The four scenarios:

| Dir | Outcome |
|---|---|
| `rekor2-dsse-happy-path` | valid; must verify |
| `rekor2-dsse-mismatch-sig_fail` | same cert, same PAE, but `envelope.sig != tlog.sig` (a second valid signature of the same PAE was swapped in) — exercises §4.4.1 #2 (signature-equality) |
| `rekor2-dsse-mismatch-envelope_fail` | same cert, envelope swapped for one covering a different payload, so `Hash(PAE(envelope)) != tlog.digest` — exercises §4.4.1 #1 (digest-equality) |
| `rekor2-dsse-invalid-sig_fail` | `envelope.sig` was made by an unrelated key and does not verify under the bundle's certificate |

In every `_fail` dir the RFC3161 TSA timestamp's `messageImprint` is over the *swapped-in* envelope signature (not the original tlog signature), so the timestamp check passes and the bundle fails only for the labelled reason.

All bundles use the `sigstore-conformance/extremely-dangerous-public-oidc-beacon` workflow identity (the suite's default in [`test/client.py`](https://github.com/sigstore/sigstore-conformance/blob/main/test/client.py)), so no per-dir `identity`/`issuer` overrides are needed.

Related: https://github.com/sigstore/architecture-docs/pull/63


loosebazooka:
Might be worth splitting into v1 and v2 changes. At least we can get the v2 changes in?

loosebazooka:
I'll test the v2 changes against the sigstore-java impl for now though.

Hayden-IO:
I thought in the sigstore-go meeting we had discussed doing this only for Rekor v2?

+1 to splitting these. I don't think we should pursue this change in Rekor v1 if it's a breaking change to verification. We will have to test this and conformance is the easiest way to do so. But I assume it will be a breaking change because clients won't know to construct the PAE as the artifact, either when comparing it to the hash digest in the canonicalized body or if the client reconstructed the leaf hash themselves.

I want to avoid having a divergent signing path with a flag to control how signing is done for v1, that's just more technical debt we have to keep track of.

loosebazooka:
the v2 tests are working as expected in sigstore-java (on my branch).

I think we need to get this into sigstore-python first maybe? (or xfail on the self test... we've done this before as long as there was confirmation that they work)? @jku

loosebazooka:
also cosign-nightly is gonna be mad if it can't handle these new tests.

loosebazooka:
otherwise lgtm

Hayden-IO:
Can we create a PR to xfail these tests with Cosign before this lands?

codysoyland:
I have a PR for sigstore-python here: https://github.com/sigstore/sigstore-python/pull/1776

jku:
I'll have a look at sigstore-python today but I have not problem with xfails being added for  the selftest client (as long as someone has tested the cases with another compliant client).

jku:
I can make a release of the current state of conformance if that makes it easier for clients. We can make another one when this merges

jku:
we can upgrade the selftest client even without a sigstore-python releases:
`sigstore @ git+https://github.com/sigstore/sigstore-python.git@603eeb3d4f6064917926fdaacc832790cbc9a609` in selftest-requirements.txt should work

(I see we've forgotten to update version in there the last time something like this happened, oops)

jku:
Is this ready to go? I think we could merge this so clients could easily test against main branch -- we could still hold off releasing new conformance until the main clients have their implementations ready.

codysoyland:
> Is this ready to go? I think we could merge this so clients could easily test against main branch -- we could still hold off releasing new conformance until the main clients have their implementations ready.

Yeah, it should be good to go now. I just ran it against my local branch of sigstore-go again and it passes.

loosebazooka (approved):

jku (approved):
lgtm, thanks for all your work on this

PR #372, opened 2026-05-15
Fix and add SET failure tests

aaronlew02:
#### Summary

This change replaces the bundle in `integrated-time-in-future_fail` with one that targets the intended error (originally, that test failed due to an invalid SET signature) and adds a new `set-invalid-signature_fail` test to continue testing an invalid SET signature in isolation.


loosebazooka:
it looks like the local failures I'm seeing are from a cert id/issues mismatch? If we're using the new token, is it inherently supported? Or do we have to make some code modifications?

So it looks like you need to add [issuer and identity](https://github.com/sigstore/sigstore-conformance/blob/main/test/assets/bundle-verify/README.md) files to each new test dir. Or recreate these tests with the old conformance oidc token

aaronlew02:
I've added issuer and identity files for `integrated-time-in-future_fail` and replaced the malformed SET in `set-invalid-signature_fail` with a parseable SET from a different bundle to better target the intended case.

loosebazooka:
Okay local testing seems to show the right errors. Nice work.

loosebazooka (approved):

jku (approved):

PR #376, opened 2026-06-02
Create RELEASE.md

loosebazooka:


jku (dismissed):
I guess we could add step 2 ( that I didn't do for current release yet): _Update versions in examples in the README_

jku (approved):

Issue #382, opened 2026-06-11
test for signing dsse-as-hashedrekord


jku:
We already have verify tests for dsse-as-hashedrekord but could add a full sign-and-verify test in `test_bundle.py` (a mash up of test_sign_verify_rekor2 and test_sign_verify_dsse)

PR #383, opened 2026-06-11
Conformance report: Add DSSE-as-Hashedrekord checkbox

jku:
Also combine rekorv2 verify and sign into a single check

The new check only depends on verifying DSSE-as-Hashedrekord as it turns out we don't test signing at the moment (#382).

### DIFF
--- a/.github/scripts/generate_client_report.py
+++ b/.github/scripts/generate_client_report.py
@@ -19,6 +19,7 @@ class Result:
     skipped: int = -1
     rekor2_verify: bool = False
     rekor2_sign: bool = False
+    dsse_hashedrekord: bool = False
     managed_keys: bool = False
     client_sha: str = ""
     client_sha_url: str = ""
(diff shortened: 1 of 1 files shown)


.github/scripts/generate_client_report.py
```diff
@@ -85,8 +88,8 @@ def _generate_html(results: list[Result]):
                     <th>Failed</th>
                     <th>Skipped</th>
                     <th>Xfailed</th>
-                    <th>Rekor v2 verify</th>
-                    <th>Rekor v2 sign</th>
+                    <th>Rekor v2</th>
+                    <th>DSSE-as-Hashedrekord</th>
```
loosebazooka, line 92: Should we specify it's for rekor v2?

loosebazooka (approved):

.github/scripts/generate_client_report.py
jku: Maybe... but I guess you'd have to look at the actual test to see what these are anyway

PR #393, opened 2026-07-08
Add SCT duplicate-intermediate-name verification test

wolfv:
## What

Adds a `bundle-verify` fixture, `sct-duplicate-intermediate-name`, that exercises SCT issuer resolution when the trusted root contains two Fulcio intermediates with the **same Subject DN but different keys**.

## Why

Fulcio's SCT is verified against `issuer_key_hash = SHA-256(issuer SubjectPublicKeyInfo)` (RFC 6962 precert). A client must resolve *which* intermediate issued the signing certificate and hash *that* key. A client that resolves the issuer by **Subject name alone** breaks the moment the trusted root holds two identically-named intermediates — it may hash the wrong key and reject a valid SCT.

This happened in production: during the July 2026 Fulcio multi-region migration, staging began serving a second, identically-named intermediate, which broke name-only SCT issuer resolution in sigstore-rust (fixed in sigstore/sigstore-rust#154). There was no conformance test covering this shape.

## How the fixture works

- **Bundle:** the standard beacon-identity `happy-path` bundle, so it passes the suite's fixed identity/issuer policy unchanged — issuer resolution for SCT is the only variable under test.
- **Trusted root:** the real public-good production root, with a **decoy** certificate authority prepended. The decoy's intermediate shares the genuine intermediate's Subject DN (`CN=sigstore-intermediate,O=sigstore.dev`) byte-for-byte (PrintableString-encoded) but carries an unrelated key, and sorts before the genuine CA.

A correct client sources the issuer key from the cryptographically verified chain and verifies the SCT; a name-only resolver hashes the decoy's key and fails with an invalid SCT signature. It is a positive test (no `_fail` suffix): a conformant client must succeed.

`make_fixture.py` is included for reproducibility; the committed `trusted_root.json` is canonical.

## Validation

| Client | Result |
|---|---|
| sigstore-python 4.1.0 (reference) | ✅ pass |
| sigstore-rust (post-fix) | ✅ pass |
| sigstore-rust (pre-fix) | ❌ fail — `SCT signature verification failed: ECDSA P-256 SHA-256 signature invalid` |

Run through the pytest harness (both PATH and DIGEST variants) against a conformant client and it passes; against the pre-fix client it fails — i.e. the test catches the regression.


jku:
Makes sense to me.

> Opened as a draft to let the full client matrix (sigstore-go, sigstore-js, cosign, …) run in CI and confirm the fixture is implementation-agnostic before merge.

This we don't have: conformance itself currently only tests with the selftest client (python). Maybe we should test against more than one client but I doubt we want the whole set in CI here -- writing conformance tests is already a bit unappealing, having to debug N clients would make it worse.

Leont:
Yes this sounds like a good test to me.

PR #402, opened 2026-08-14
Add canonical Rekor intoto verification fixture

wolfv:
Add a positive managed-key Rekor v1 `intoto/0.0.2` fixture using the canonical log-entry representation: `payloadHash` is present and the full DSSE payload is omitted. The fixture uses a one-leaf custom transparency log and a SHA-256 subject so unrelated certificate-time and digest-algorithm checks cannot mask canonical-body parsing.

The sigstore-python self-test is marked xfail because its wrapper does not currently expose managed-key verification.

Validation against sigstore-rust at audited commit `6d268a5d8005e5c3e91a44c20fcfd557f514e668` fails for artifact and digest verification with the expected `missing field payload` parser error.


wolfv:
Validated against the report target, sigstore-rust `6d268a5d8005e5c3e91a44c20fcfd557f514e668` (`cargo build --release -p sigstore-conformance`). Both PATH and DIGEST cases fail as intended for this positive fixture, with `failed to parse intoto v0.0.2 body: missing field payload`.

PR #403, opened 2026-08-14
Test TSA validity binding

wolfv:
Add a negative Rekor v2 fixture where the timestamp signer is outside its configured validity period while an unrelated, certificate-less TSA entry covers the timestamp. Verification must not borrow temporal authorization from another TSA.

Checks:
- sigstore-python rejects both artifact and digest verification
- the affected sigstore-rust verifier unexpectedly accepts both, demonstrating the fixture catches the issue


wolfv:
Validated against the report target, sigstore-rust `6d268a5d8005e5c3e91a44c20fcfd557f514e668`. Both PATH and DIGEST negative cases fail the conformance test because the older verifier unexpectedly succeeds.

Hayden-IO:
To confirm - this is testing that a TSA where the issued timestamp is outside the validity period of the TSA (per the trusted root) can't be used to verify a short-lived certificate?

If so, should we have a similar test for Rekor v1 timestamps, where somehow an SET was issued outside of the validity window of the log?

PR #404, opened 2026-08-14
Test all timestamps against certificate validity

wolfv:
Add a negative Rekor v2 fixture containing two cryptographically valid RFC 3161 timestamps over the same signature. The first is within the signing certificate validity period; the second is after expiration. Clients must reject the bundle rather than selecting only the earliest timestamp.

Checks:
- sigstore-python rejects artifact and digest verification
- the affected sigstore-rust verifier unexpectedly accepts both, demonstrating the fixture catches the issue


wolfv:
Validated against the report target, sigstore-rust `6d268a5d8005e5c3e91a44c20fcfd557f514e668`. Both PATH and DIGEST negative cases fail the conformance test because the older verifier unexpectedly succeeds.

Hayden-IO:
Why should this fail? Verification should be as follows:

* clients verify as many sources of time as possible. Currently that’s signed rfc3161 timestamps or Rekor signed timestamps.
* clients should be configured with a verification policy with a threshold for min number of verified timestamps. Clients should verify the number of validated timestamps against this policy 
* clients should verify that at least one timestamp falls in the window of validity for the signing certificate

If one of the valid timestamps occurred outside of the validity window, I think clients should simply skip over it. We could also apply another threshold to the number of timestamps that have to be in the window of validity of the cert, although I see less value in that.  

We may need to clean up the spec around this point. I was looking over it yesterday and it’s not well defined on time.

cc @loosebazooka since we were chatting about this

Hayden-IO:
Uh oh, i might be inconsistent, Sigstore go does whats tested here….

https://github.com/sigstore/sigstore-go/blob/22d3691c7b8e0c5530fae3c05577690bfef5cd00/pkg/verify/signed_entity.go#L692

wolfv:
Apparently `sigstore-python` also agrees with `sigstore-go` and rejects this, right?

Hayden-IO:
Well, if all the other clients are in agreement, then we can merge this and start up a separate thread to discuss timestamp verification semantics.

There's a larger question here around _weird_ bundles, meaning bundles that could have only been crafted by custom tooling or manually. Do we err on the side of acceptance - if at least a subset of the bundle is verifiable per the spec, allow it - or should clients reject - only if all permutations of each component of the bundle are verifiable per the spec, allow it?

jku:
> Apparently sigstore-python also agrees with sigstore-go and rejects this, right?

I have a vague recollection that we considered this not a spec requirement (because like hayden says it doesn't seem to be clearly required) but a client safety decision: There's something fishy about this bundle and we're erring on side of caution until there's a reason not to.

I don't feel strongly about it and it's possible that decisions like this are not good for the ecosystem (as they could lead to more and more bundles being somehow problematic) but they are a lot easier to make while coding/reviewing clients than the permissive-by-default decisions...

PR #405, opened 2026-08-14
Test in-toto statement type validation

wolfv:
Add two cryptographically valid managed-key Rekor v2 DSSE bundles whose in-toto statements have either an unknown `_type` or no `_type`. Verifiers must reject both rather than interpreting them as Statement v1.

The fixtures use a one-leaf custom transparency log, so certificate/SCT validation cannot mask statement-type behavior.

Checks:
- sigstore-python rejects all four artifact/digest cases
- sigstore-rust at audited commit `6d268a5d8005e5c3e91a44c20fcfd557f514e668` unexpectedly accepts both unknown-type cases
- current affected sigstore-rust build also unexpectedly accepts both unknown-type cases


wolfv:
Validated against the report target, sigstore-rust `6d268a5d8005e5c3e91a44c20fcfd557f514e668`. The unknown `_type` fixture fails for both PATH and DIGEST because the older verifier unexpectedly succeeds. The missing `_type` controls pass because that verifier already rejects a missing field. Overall: 2 failed, 2 passed.

Hayden-IO (dismissed):
As an aside, I wonder if we could have a separate set of tests for intoto and dsse conformance? Maybe something to donate to those projects?

PR #406, opened 2026-08-14
Test trust-root validity interval semantics

wolfv:
Add trust-root fixtures for the specified `TimeRange` semantics:

- tlog key end equal to integrated time succeeds
- TSA end equal to timestamp succeeds
- a tlog validity interval missing its required start fails

`TimeRange` is documented as a closed interval. The sigstore-python self-test is therefore marked xfail for the TSA end-boundary case it currently rejects.

Checks:
- sigstore-python: four pass, two expected failures
- current sigstore-rust conformance client: all six pass


wolfv:
Validated against the report target, sigstore-rust `6d268a5d8005e5c3e91a44c20fcfd557f514e668`. The missing-start fixture fails for both PATH and DIGEST because the older verifier unexpectedly succeeds; the four closed-boundary controls pass. Overall: 2 failed, 4 passed.

jku:
Are the custom trusted roots modified in some way?

Would be nice to have that documented in the READMEs (in case they need to be recreated)

wolfv:
Good point. I added the source fixture, exact field transformation, key ID, and boundary timestamp to each README. I skipped a generator script because these are three one-field JSON edits; a script would add more maintenance than reproducibility here.

jku (approved):
lgtm, thanks

jku:
In general, I think we can merge these (if we agree on the ideas), then after the merges can do another test run on a 3rd client just to see if there's anything unexpected before releasing

